PR #12, opened 2026-02-14
Stop closing terminal after running

TheUntraceable:
Hi, I tried using this as a fellow Windows user in Powershell but I couldn't get it to work. It annoyed me that I couldn't see the command to copy the save in PATH forever because the terminal closes because I didn't want to install the skill (I was in the root users folder in a new terminal). I think this might be a good change, feel free to reject if you disagree!

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -195,7 +195,6 @@ function Main {
     }
 
     Write-Host ""
-    exit 0
 }
 
 Main


TheUntraceable:
PS: Where are your docs code? please mention your NPX package instead of having to install it natively. Npx saved me a lot of time.

TheUntraceable:
What the fuck was I thinking that doesn't stop the terminal that's fine. please ignore my stupidity I swear I'm not AI slop

nc9:
haha - thanks for the effort. I want to get this resolved, I don't have access to a windows machine - but around 20% of users are on windows from what I see - but they may be using the linux subsystem. I may have to spin up a vm to test it

TheUntraceable:
Feel free to ask me, more than happy to run through a test install for windows. I will mention when it prompts me to install the skill I decline it and it restarts the terminal (havent tested what happens if i accept). I just used the npx package , ~~maybe deprecate the windows shit if the npx global install works fine? Up to you~~

EDIT: I failed to consider this isn't a TS Dev specific thing; bad to assume everyone has got node installed.

PR #70, opened 2026-08-03
fix(audit-engine): populate contactLinks on the adapter parse path

nc9:
Follow-up to the cross-page NAP drift work. A review pass after that merged found the new `ParsedPage.contactLinks` field is populated by only one of the two parsers.

## The bug

There are two functions that produce a `ParsedPage`:

- `parsePage` (`@squirrelscan/parser`)
- `parseHtmlForRules` (`packages/audit-engine/src/adapter.ts`), which is what `parsePageRecord` and the page-rule workers actually run

`contactLinks` was added to the first only. So in a real audit the field came back undefined, and `extractNapSignal` reads nothing from it: every phone number declared as a `tel:` link was invisible, along with the `contact-info` contact-page check. Only JSON-LD `telephone` still reached the rule. Every parser-level test passed the whole time, because they call the other parser.

## The fix

`parseHtmlForRules` now calls `extractContactLinks(doc)` (newly exported from the parser barrel), and a regression test asserts the two parse paths produce identical `contactLinks` and that a NAP signal survives the adapter path. That test is the actual guard: a field added to one of two `ParsedPage` producers is invisible without it.

## Also in here (comments only)

Two behaviours worth stating in the source rather than leaving for the next reader to rediscover:

- `napPhoneKey` compares the trailing 7 digits, so two different numbers sharing a subscriber tail collapse into one key. That under-reports a real conflict as format drift and never the reverse, which is the direction that matters.
- The streaming path reads zero feature rows as an empty crawl. A crawl where every page is non-auditable also yields zero rows while the legacy universe still appends them, so the skip is worded differently. That is a property of the streaming universe shared by every site rule on this seam, not something specific to this rule.

## Verification

Root `bun test`: 3788 pass, 0 fail. `bun run test:engine`: 49 pass, 0 fail, with the canonical golden unchanged (findings 95711, `perRuleTally.length` 264, healthScore 48). Typecheck, lint and format check clean.

### DIFF
--- a/apps/cli/src/audit/adapter.ts
+++ b/apps/cli/src/audit/adapter.ts
@@ -9,6 +9,7 @@ import {
 } from "@squirrelscan/audit-engine";
 import {
   detectPageType,
+  extractContactLinks,
   extractContent,
   extractHeadings,
   extractImages,
@@ -245,6 +246,11 @@ export function parsePageRecord(page: PageRecord): ParsedPage | null {
           rel: l.rel,
           isNofollow: l.isNofollow,
         })),
+        // tel:/mailto: anchors, which `extractLinks` drops as non-crawlable.
+        // This is the THIRD ParsedPage producer (see @squirrelscan/parser's
+        // `parsePage` and audit-engine's `parseHtmlForRules`); a field set on
+        // only some of them is undefined at runtime with every test still green.
+        contactLinks: extractContactLinks(doc),
         images: images.map((i) => ({
           src: i.src,
           alt: i.alt,
--- a/apps/cli/tests/audit/parsed-page-producer-parity.test.ts
+++ b/apps/cli/tests/audit/parsed-page-producer-parity.test.ts
@@ -1,0 +1,105 @@
+// All THREE ParsedPage producers must agree.
+//
+// A `ParsedPage` is built in three separate places:
+//   1. `parsePage`          — @squirrelscan/parser (the canonical one)
+//   2. `parseHtmlForRules`  — @squirrelscan/audit-engine (cloud + page-rule workers)
+//   3. `parsePageRecord`    — apps/cli/src/audit/adapter.ts (this CLI's own copy)
+//
+// Every ParsedPage field is optional or independently defaulted, so adding one to
+// a subset of them COMPILES, and each producer's own tests keep passing. The field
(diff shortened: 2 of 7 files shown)


nc9:
Answering the forked-adapter question: **there is a third producer, and this PR was incomplete.** Pushed a second commit.

`apps/cli/src/audit/adapter.ts` `parsePageRecord` is not a delegation, it is a hand-rolled copy of `parseHtmlForRules` that builds the `ParsedPage` object literal inline. It was missing `contactLinks`, so the CLI audit path (the shipped `squirrel` binary) still dropped every `tel:`-declared phone number.

It is live, not dead code, via two callers:
- `apps/cli/src/reports/reconstruct.ts:254`
- the re-parse fallback in `buildSiteContext` when a page has no stored `parsedData`

`apps/cli/src/parse/html.ts` is a clean re-export of the canonical parser, so it is not a fourth.

## Enumerating the producers

To be sure three is the real number rather than the number I happened to find, I keyed on a field every full `ParsedPage` construction sets:

```
$ grep -rn "visibleDatePublished:" --include="*.ts" apps/ packages/ | grep -v /tests/
apps/cli/src/audit/adapter.ts:266        <- parsePageRecord   (this commit)
packages/audit-engine/src/adapter.ts:380 <- parseHtmlForRules (previous commit)
packages/parser/src/html.ts:401          <- parsePage         (already merged)
```

Exactly three, all now populated.

## The guard

I replaced the per-producer assertions with one parity test, `apps/cli/tests/audit/parsed-page-producer-parity.test.ts`, in the one package that can import all three. It asserts they produce identical `contactLinks` for the same HTML, and that none of them leak contacts into the crawlable `links` graph.

Testing a producer in isolation is precisely what let this through twice. Each producer's own suite stays green while the field is `undefined` wherever the un-updated one ran, and because every `ParsedPage` field is optional it all still compiles. Parity across producers is the only assertion that fails.

I verified the guard works rather than assuming: removing the one adapter line turns 2 of the 4 tests red, and restoring it turns them green.

## Verification

Root `bun test`: 3792 pass, 0 fail. CLI typecheck, lint and format clean. Still count-neutral, so no golden pin moves.

Holding for #69 as instructed, then I will rebase and re-run. Not merging.

PR #82, opened 2026-08-04
fix(rules): report a missing robots.txt as an error

nc9:
`crawl/robots-txt` declares meta severity `error`, but the missing-file check emitted `status: "warn"`. Effective severity derives from what actually happened (a rule whose checks only warn is reported as a warning), so a site with no robots.txt has always rendered as **Warning**, and sorted into the warning block below recommendations.

We already treat a missing robots.txt as critical: it carries a -15% score penalty. This makes the badge match.

**Changes**
- `packages/rules/src/crawl/robots-txt.ts` — missing robots.txt is now `fail`, not `warn`.
- `packages/audit-engine/src/scoring.ts` — the `PENALTY_NO_ROBOTS_TXT` (-15%) multiplier matched `status === "warn"` only. Now matches `fail` **or** `warn`, so the penalty survives the new status and still applies when rescoring reports stored before this change.
- `apps/cli/tests/scoring.test.ts` — test that the penalty applies on a failed `robots-txt-exists` check.

**Effect on scores:** a missing robots.txt now contributes a failed check rather than a warning, so it weighs more in the issue-density penalty (fail units are 2x warn units) on top of the unchanged -15% multiplier.

### DIFF
--- a/apps/cli/tests/scoring.test.ts
+++ b/apps/cli/tests/scoring.test.ts
@@ -254,6 +254,72 @@ describe("calculateHealthScore", () => {
     expect(score.overall).toBe(37);
   });
 
+  // The rule emits "fail" for a missing robots.txt (#1535); "warn" above is the
+  // legacy status stored in pre-#1535 reports. Both must keep the penalty —
+  // matching only one silently drops -15% from every score.
+  test("applies penalty for missing robots.txt when the check failed", () => {
+    const mockRobotsMeta = {
+      id: "crawl/robots-txt",
+      name: "Robots.txt",
+      description: "Robots.txt check",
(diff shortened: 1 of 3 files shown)


nc9:
Superseded by #94, which fixes the same robots.txt severity issue against current `main` and also covers the report-diff path (`apps/cli/src/reports/diff/index.ts`) that this PR does not.

This branch is a week behind a `main` that has since moved considerably (269 rules, dependency overrides, two new site-norm rules), so rebasing it costs more than taking the newer implementation. Closing to avoid two competing fixes for one rule.

No work is lost: #94 is a strict superset of the changes here.

PR #88, opened 2026-08-12
docs(projects): use FileTree for remaining hierarchy blocks

nc9:
Converts the two remaining hierarchy blocks on `/projects` to Tangly's `FileTree`
component:

- **"Viewing project data"** — a one-row Markdown table describing `project.db`
  under `~/.squirrel/projects/<project-name>/` becomes a tree, with the
  description as a trailing comment and the templated segment highlighted.
- **"Project storage location"** — a bare fenced block holding a single path
  plus a sentence of prose becomes a tree showing the root with two example
  projects and a `...` placeholder.

The templated segment is escaped as `&lt;project-name&gt;`; unescaped it parses
as a JSX tag and breaks the build.

The page's main `~/.squirrel/projects/` block was already converted in `242cc04`.

### Verification

- `tangly check` clean — 370 pages, 1 orphan (`agent-setup/prompt.mdx`, pre-existing).
- All three trees on `/projects` confirmed rendering in a dev server: directory
  and file icons, trailing comments, highlight.
- `tangly build` fails for an unrelated pre-existing reason (a stale Astro
  `content-modules.mjs` in the local Tangly checkout resolving against a deleted
  temp dir). Confirmed pre-existing by stashing this change and rebuilding —
  identical failure.

### Note on the branch

This branch is stacked on `docs/ste-house-style`, which was local-only and
unpushed, so this PR carries **7 commits**: six pre-existing docs commits from
that branch plus this one. Only `4bf7295` is mine. Review the last commit, or
rebase the rest onto `main` separately.

### Related

A full end-to-end docs flow review accompanies this change at
`plans/docs-flow-review.md`. That path is gitignored in this repo, so the
document is not in this PR — it is on disk in the working copy. Headline
findings: `--format json` has four mutually incompatible documented shapes and
the docs' own `jq` examples return `null` against each other; `configuration/output.mdx`
documents a `sarif` format that does not exist; `tangly check` fails against the
pinned `tangly@0.2.4` because `docs.json` targets an unreleased theme.


nc9:
Closing: opened by mistake by an agent working from a mis-dispatched task; it also published commits from an in-progress local branch that weren't ready. All work is preserved locally and will be re-raised deliberately.

PR #95, opened 2026-08-13
fix(installer): capture self install output and fix Windows symlink EPERM

nc9:
Windows PowerShell installs fail at `squirrel self install`, and the installer only reports the exit code — so the failure is undiagnosable (private tracker squirrelscan/repo#1538, Sentry `INSTALLER-WORKER-A`: 3 events since 2026-08-03, windows/x64/ps1, step `self_install`, exit 1).

## What was wrong

**No diagnostics.** Both installers ran the binary bare (`& $binaryPath self install` / `"$tmpdir/squirrel" self install`), so the error report carried an exit code and nothing else.

**The actual failure.** `runSelfInstall` ends with `fs.symlinkSync(binaryPath, symlinkPath)`. Windows grants `SeCreateSymbolicLinkPrivilege` only to elevated processes and Developer Mode accounts, so this throws `EPERM` for a normal user — every stock Windows install. CI missed it twice over: the GitHub Windows runner is elevated (symlinks succeed), and `test-install.yml` downloads the **published** `install.ps1` from `main` rather than the one under test.

## Changes

- **Capture the output.** `Invoke-CapturedCommand` (ps1) and `tee` + `PIPESTATUS[0]` (sh) echo the command's stdout/stderr live *and* keep it. On failure a scrubbed, bounded tail ships as a new `error_output` field; report version bumps to `2`. Printable ASCII only, `$HOME`/`%USERPROFILE%` → `~`, 1000 chars, tail kept (the failure is at the end). The worker re-clamps and redacts server-side.
- **Fix the root cause.** `linkBinary()` falls back to `copyFileSync` on win32 when the symlink is refused. POSIX still symlinks, and an EPERM there still surfaces rather than silently becoming a copy.
- **Make a copy install a first-class install.** `isManagedInstall()` accepts the managed Windows bin path (otherwise `self update` would refuse and auto-update would be off), `updateSymlink()` renames a running `.exe` aside before the swap (Windows can rename a loaded exe, not overwrite it) and sweeps the leftovers, and `self doctor` treats a Windows file copy as a pass.
- **CI parses install.ps1** with `pwsh` on the ubuntu runner, plus `bash -n install.sh`. Nothing here parsed that file before users did.

## Notes

- Piping the child's output means it is no longer a TTY, so `self install`'s output is now uncoloured during install. Its content is unchanged.
- `install.sh` takes the last 40 **lines**, not bytes: a byte cut can split a `$HOME` path so the scrubber no longer matches it.
- The consuming `/error` endpoint change (accept + surface `error_output` on the Sentry event) lands in the private monorepo alongside this.

## Testing

- `scripts/install-contract.test.ts`: contract assertions for both scripts, plus a behavioural test that sources install.sh's reporting preamble and asserts the JSON `report_error` actually POSTs (bounded to 1000 chars, tail kept, `$HOME` scrubbed).
- `apps/cli/tests/self/link-binary.test.ts`: symlink on posix, copy fallback on windows-denied, no fallback on posix EPERM.
- `bun test apps/cli/tests/self scripts/` → 270 pass; typecheck/lint/format clean.
- PowerShell could not be executed locally (no pwsh on the dev box); the new CI job parses it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,31 @@ jobs:
       - uses: ./.github/actions/setup-bun
       - run: bun test
 
+  installer-scripts:
+    name: Installer script syntax
+    needs: quality
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      # install.ps1 only ever runs on Windows machines, and the install test
+      # workflow fetches the PUBLISHED script from main — so nothing here ever
+      # parsed a change to it before users did. pwsh ships on the ubuntu runner
+      # and exposes the same parser Windows PowerShell uses.
+      - name: Parse install.ps1
+        shell: pwsh
+        run: |
+          $errors = $null
+          [System.Management.Automation.Language.Parser]::ParseFile(
+            (Resolve-Path './install.ps1').Path, [ref]$null, [ref]$errors) | Out-Null
+          if ($errors) {
+            $errors | ForEach-Object { Write-Host $_ }
+            exit 1
+          }
+          Write-Host 'install.ps1 parsed clean'
+      - name: Parse install.sh
+        run: bash -n install.sh
+
   dependency-audit:
     name: Dependency audit
     needs: quality
--- a/apps/cli/src/controllers/self/install.ts
+++ b/apps/cli/src/controllers/self/install.ts
@@ -2,7 +2,6 @@ import {
   existsSync,
   mkdirSync,
   copyFileSync,
-  symlinkSync,
   unlinkSync,
   chmodSync,
 } from "node:fs";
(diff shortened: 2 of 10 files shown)


nc9:
## CodeQL alert 199 dismissed as a false positive

`js/file-system-race` (high) fired on `apps/cli/tests/self/link-binary.test.ts`. Recording the reasoning here since dismissing a high-severity alert should leave an audit trail.

**Why it is not exploitable.** The flagged pair is `lstatSync(link)` then `readFileSync(link)`, inside a unit test, on a path the test created moments earlier inside its own `mkdtempSync` directory and removes in the same `finally`. There is no second writer and no attacker-controlled path, so the TOCTOU window the rule describes cannot be reached.

**Why not restructure instead.** I first removed a genuinely redundant `existsSync(link)` — `lstatSync` throws when the path is missing, so it asserted nothing. That was a real improvement, but the rule simply moved onto the next stat-then-use pair, because it fires on the shape rather than the exposure. The two remaining assertions are both load-bearing: one proves the Windows fallback produced a real file rather than a symlink, the other proves the bytes were copied. Satisfying the analyzer would mean an `open`/`fstat`/`read` handle dance in a unit test — worse code, no security gain.

The redundant-`existsSync` cleanup is kept regardless, since it made the test slightly better on its own merits.

PR #99, opened 2026-08-13
feat(rules): add social/asset-divergence, site chrome that disagrees across pages

nc9:
New site-wide rule `social/asset-divergence`: the favicon, the `theme-color` and the fallback OG image all come from one layout, so a pocket of pages carrying different ones is a stale template rather than a per-page decision. Nothing about such a page looks wrong on its own, which is why no per-page rule can see it.

Private-repo issue: squirrelscan/repo#1371.

## What it does

Three axes, each judged against the site's own modal value, only the minority form reported:

| Axis | Read from |
|---|---|
| favicon | first `<link rel="icon">` href, resolved absolute (apple-touch/mask icons never stand in) |
| theme-color | `<meta name="theme-color">` content (an unscoped declaration wins over a light/dark `media` pair) |
| default OG image | `og:image`, on the pages that should share one |

## Precision guards

- 10-page crawl floor; 10-page per-axis sample floor; 70% modal agreement, else the axis is not judged and the rule emits one `skipped` check.
- A page that declares nothing never votes and is never a deviant — a missing favicon belongs to `core/favicon`, a missing share image to the Open Graph rules.
- **The OG axis excludes per-page images two ways**: article/product/recipe/event/profile/media page types, *and* any `og:image` carried by exactly one page. The second one came out of replaying the rule over stored real crawls — sites ship bespoke preview images on pages the classifier calls "unknown", which the page-type exemption alone would not cover. A fallback is by definition shared; a stale template always brings its whole pocket of pages with it.

## Dual path

`ctx.siteQuery` streams the three scalars off `page_features`; the legacy `ctx.site.pages` path re-derives them through the SAME exported extractor `extractPageFeatures` calls. One accumulator, one check builder, every list sorted before it is emitted — byte-identical by construction, following `content/duplicate-title` (#1021/#1022).

Storage: three additive `PageFeatureRow` columns (`favicon_href`, `theme_color`, `og_image`) in the fresh-install schema AND an append-only schema-version 21 migration. Local sqlite only, no prod migration, existing versions untouched.

## Scoring density

One site-wide check, weight 3, items capped at 10 groups / 10 urls, warn below a 20% deviant share and fail at or above it. On the canonical 518-page fixture the health score is **unmoved at 48**; `findings` moves 98216 → 98217 (one site-scoped check) and `perRuleTally` 272 → 273.

## Tests

- `packages/rules/tests/asset-divergence.test.ts` — 18 tests: pass / warn / fail / skipped (under-floor, too-few-declaring, no-clear-norm), the no-false-positive fixtures (articles + products with their own share image over a uniform default; a bespoke one-off image on an untyped page), and the extraction edge cases. Every fixture runs BOTH paths and asserts `toEqual` **and** `JSON.stringify` equality.
- `packages/audit-engine/tests/site-query-asset-divergence-golden.test.ts` — the full chain (`parsePage` → `extractPageFeatures` → v21 columns → `createSiteQuery` → streaming branch) against one seeded fixture, with the siteQuery run given empty `site.pages`; plus a pin that both parse paths (`parsePage` and `parseHtmlForRules`) yield the same signal.
- `packages/crawler/tests/page-features-store.test.ts` — round-trip of the three columns, nulls, and the v21 columns on BOTH the fresh-install and migration paths.

## Real-corpus smoke

Replayed over 7 stored crawls (sweetdeals.com.au, openelectricity.org.au, nikcub.me, gohugo.io, httpd.apache.org, digitalpages.com.au, sweetgreen.com): **zero findings, zero false positives** — each correctly `skipped` for want of a norm. openelectricity is the instructive one: 100 pages declare an `og:image` and 95 of them are unique to their page, so nothing there is a deviant.

## Checks

- root `bun test`: 4132 pass, 0 fail
- `bun run test:engine` (golden gates incl. the 518-page v1↔v2 canonical merge gate): 52 pass, 0 fail
- `bun run typecheck`, `bun run lint`, `bun run format:check`, `docs/ bun run check`: clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Combine your coding agent with a deterministic and extensible audit tool.
 
 ## Features
 
-- **277 Rules, 21 Categories** - Comprehensive coverage across SEO, accessibility, performance, and security
+- **278 Rules, 21 Categories** - Comprehensive coverage across SEO, accessibility, performance, and security
 - **Fast crawler** - Highly optimized memory efficient crawler 
 - **Agent Experience** - Audit agent experience to assist agents in using your site
 - **Security Audit** - Detect phishing kits, leaked credentials, and more 
@@ -49,7 +49,7 @@ Ordered by how much a failure usually costs you, not by how many rules each one 
 | Structured Data | 12 | JSON-LD validity and rich-result eligibility for Article, Product, FAQ, Review, Breadcrumb, Organization, LocalBusiness, Video and site search, plus rating markup that is not about the page it sits on |
 | Accessibility | 61 | WCAG coverage: ARIA roles and names, form labels and autocomplete tokens, colour contrast, heading order, landmarks, tables and lists, focus visibility, touch targets, captions |
 | Mobile | 6 | Viewport configuration, tap-target size, legible font sizes, horizontal scroll, blocked zoom, intrusive interstitials |
-| Social Media | 4 | Open Graph and Twitter Card completeness, image dimensions, canonical URL match, social profile links |
+| Social Media | 5 | Open Graph and Twitter Card completeness, image dimensions, canonical URL match, social profile links, site-chrome assets that disagree across pages |
 | URL Structure | 9 | Length, casing, hyphenation, stop words, query parameters, special characters, trailing-slash consistency, site-wide convention consistency |
 | E-E-A-T | 15 | Experience, expertise, authority and trust signals: author bylines and credentials, about and contact pages, citations, editorial policy, disclaimers, YMYL detection |
 | Legal Compliance | 4 | Privacy policy, terms of service, real cookie-consent machinery, subprocessor disclosure |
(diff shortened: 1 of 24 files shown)


nc9:
Rebased onto current `main` (after #121 landed). One conflict was substantive.

### Schema version collision — the real one

Both this branch and #115 claimed `SCHEMA_VERSION = 21`: #115 for `sitemaps.is_news_sitemap`, this branch for the `page_features` chrome scalars. Resolved by keeping 21 for #115 and renumbering this migration to **22**. They are independent `ALTER`s on different tables, so ordering is moot.

Worth naming because a "clean" resolution here would have been silent: the comment on `PAGES_ALTER_COLUMNS` documents this exact failure from #875/#839, where a DB recorded at the current version skips a migration **forever** and every write throws a missing-column error. Migration replay is safe: the runner swallows `duplicate column name`, so the wind-back test that resets to v20 now replays 21 and 22 without issue.

Also switched two pinned `toBe(21)` assertions to the exported `SCHEMA_VERSION`, which is what main added that export for.

### Rule counts

Recomputed with `loadAllRules()` rather than by arithmetic:

```
TOTAL: 278   CATEGORIES: 21   social: 5   links: 15
```

All 21 rows of both category tables were cross-checked against that. One stale count surfaced and is fixed here: `docs/rules/index.mdx` still said **Links (14 rules)** — #109 updated the README table but missed this card.

### Verification

Golden baseline lands exactly where predicted: `findings=98220 rules=278 healthScore=48` (health unmoved, as the original comment claimed). Full CI green: CLI 1728, crawler 251, rules 1375, audit-engine 278.

Note the previous all-green on this PR was from 13 Aug and did not survive contact with current main.

Issue #110, opened 2026-08-19
rules(links): anchor mismatch — inbound anchor text vs the target's title/H1/slug


nc9:
Part of #103 (V-05).

`links/anchor-text` is page-scope and looks only at a page's **outgoing** links, asking whether the
text is empty or generic. Nothing asks the question that actually matters for topicality: **does the
anchor text pointing at a page have anything to do with that page?**

Real remainders found across the audited sites, after the false positives were eliminated:

- A consultancy site — `how I work` -> `/engage`, x12, and it is the **only** anchor text that page
  ever receives.
- `squirrelscan.com` — `View all posts` -> `/blog`, x18.
- An agency site — phrasal anchors like `wire that up` -> `/services/analytics`.

Chrome anchors are template text and must be excluded, or every site fails on its own nav.

### Must-not-fire — four traps, each of which produced phantom findings

1. **Extract anchor text with a separator between child nodes.** Concatenating them yields
   `CarpentryStructural & finish carpentrySee details`, which matches nothing — **171 phantom
   mismatches on one site** from this alone.
2. **Plurals must match.** `Fridges` -> `/repairs/fridge` is a match. Compare with a stem or prefix
   rule, not set intersection.
3. **Two-character brand tokens are real tokens.** `LG` -> `/brands/lg` must match; a `length > 2`
   token filter drops the only word in the anchor.
4. **Utility pages are never mismatch targets.** `/privacy`, `/terms`, `/contact` are linked with
   boilerplate text by design.

And the counting rule, which is the subject of V-06 (#104): judge per **(source page, target)** pair,
not per link.

Depends on #109
Depends on #104

## Acceptance Criteria
- [ ] New site-scope rule `links/anchor-mismatch`, registered in the rules catalog
- [ ] For each (source page, target page) pair joined by at least one contextual link, compares the pair's combined anchor text against the target's title, H1 and slug
- [ ] Warns only when no meaningful token is shared, and reports the anchor text, the target URL and the target's title
- [ ] Chrome links are excluded from the comparison entirely
- [ ] Anchor text extraction inserts a separator between child nodes; fixture: a card whose link wraps a heading and a description resolves to space-separated tokens
- [ ] Fixture: anchor `Fridges` -> page titled `Fridge repairs` at `/repairs/fridge` — no warning
- [ ] Fixture: anchor `LG` -> `/brands/lg` — no warning
- [ ] Fixture: `/privacy` linked as `Privacy` from every page — no warning
[...]

Issue #109, opened 2026-08-19
rules(links): pages with zero CONTEXTUAL inbound links pass orphan-pages


nc9:
Part of #103 (V-04).

A link from the footer of every page on the site is not link equity — but the two rules that measure
inbound support count it the same as a link from body copy.

- `links/orphan-pages` flags pages below a raw incoming-link threshold.
- `links/weak-internal-links` flags pages with exactly one dofollow incoming link.

Both take `(url, incomingCount)` pairs (`collectOrphans`, `collectWeakPages`) with **no container
distinction**. So a page reachable only through a paginated index, or linked only from sitewide
chrome, has a healthy count and passes both while receiving no contextual equity at all.

**Evidence:** on one 45-page production blog, **28 of 45 pages have zero contextual inbound links**.
Every post is reachable only through the paginated blog index. Both existing rules pass them.

This is the one item in the epic that needs new plumbing: the link graph must record whether each link
sat inside site chrome (an ancestor `nav`, `header`, `footer` or `aside`), and the resulting per-page
contextual-inbound count has to be produced identically by the legacy `site.pages` path and the
`siteQuery` accumulator.

## Acceptance Criteria
- [ ] The link graph records, per link, whether it sat inside site chrome (ancestor `nav`/`header`/`footer`/`aside`)
- [ ] A bounded per-page scalar for contextual (non-chrome) inbound link count is available to rules from both the legacy `site.pages` path and the `siteQuery` accumulator
- [ ] Both paths produce identical counts for the same crawl, covered by a test
- [ ] New site-scope rule `links/no-contextual-inbound`, registered in the rules catalog
- [ ] Warns on pages whose contextual inbound count is 0 while their raw inbound count is above the orphan threshold — i.e. the case the existing rules cannot see
- [ ] Excludes the homepage and honours the same exclude patterns as `links/orphan-pages`
- [ ] Fixture: a page linked from the sitewide footer only — warns, and is NOT reported by `links/orphan-pages`
- [ ] Fixture: a page linked once from body copy — no warning
- [ ] Existing `links/orphan-pages` and `links/weak-internal-links` output is unchanged by this work
- [ ] Rule page added to `docs/` and `make validate` passes there
- [ ] `packages/audit-engine/tests/streaming-rules-canonical-golden.test.ts` updated: bump `v1.findings.length` and `v1.perRuleTally.length` by exactly one each, and add a comment line naming this rule as the cause, matching the convention already used for the six rules above it

[...]

Issue #108, opened 2026-08-19
rules(content): date agreement — visible byline vs schema vs URL vs Last-Modified


nc9:
Part of #103 (V-03).

A page carries up to five date signals: the visible byline, schema `datePublished`/`dateModified`, a
date in the URL or title, the sitemap `lastmod`, and the `Last-Modified` header. **Disagreement
between them is the defect**, and no rule looks for it.

Both existing date rules are presence checks:

- `content/freshness` (`packages/rules/src/content/freshness.ts`) pushes `pass` as soon as **any**
  signal exists, and returns `info` only when there are none.
- `eeat/content-dates` reports the coverage **percentage** of `datePublished` / `dateModified` across
  article pages.

**squirrelscan.com's own site passed both while broken in two opposite ways:**

- `/blog/*` showed a visible **2025** date against schema `datePublished` **2026**, and emitted no
  document schema at all — so a post's only JSON-LD was the sitewide `SoftwareApplication`.
- `/learn/*` emitted Article JSON-LD with dates and showed the reader **no date at all**.

### Must-not-fire — these cost two full re-runs to find

An internal audit pass's first version of this check produced **14 findings, of which 1 was real**.
Both causes must be fixtures here:

1. **A date belongs to the node that describes the DOCUMENT.** A sitewide `SoftwareApplication` /
   `Product` / `Organization` node carries its own `datePublished` (a release date). Taking the first
   one in the document attributed a placeholder `datePublished: 2026-01-01` to **every** blog post.
   Dates must come from document-describing types (`Article`, `BlogPosting`, `NewsArticle`, `WebPage`,
   `FAQPage`, ...); a fallback to any other node must be marked as such and discounted.
2. **A byline is a position, not a date-shaped string.** Scanning body text reads running prose and
   citations as bylines — `INP replaced FID on March 12, 2024` on a 2026 page, and
   `<a>March 12, 2024</a>` linking to web.dev. Only short lines near the top of main content count,
   and never an anchor's own text.

## Acceptance Criteria
- [ ] New page-scope rule `content/date-agreement` under `packages/rules/src/content/`, registered in the rules catalog
- [ ] Extracts up to four page-side signals: visible byline date, schema document date, URL/title date, `Last-Modified` header
- [ ] Warns when the visible byline date and the schema document date disagree by more than 1 day, naming both values
- [ ] Warns when a page emits schema document dates but shows the reader no visible date
[...]

Issue #107, opened 2026-08-19
rules(crawl): sitemap lastmod drifts from the page's own dateModified


nc9:
Part of #103 (V-02).

Separate from the churn case in V-01: here the sitemap *does* carry varied dates, they are just wrong
relative to what the page itself says. Nothing correlates the two today.

**Two production sites, and the two directions were different bugs:**

- A personal blog — sitemap lastmod up to **5,589 days ahead** of the page's own `dateModified`.
  Cause: lastmod took the build date while the page rendered the post's real 2011-2016 date.
- A consultancy site — sitemap lastmod **433 days behind** `dateModified` on revised posts. Cause: the
  sitemap generator resolved `publishedAt ?? updatedAt` while the route rendered
  `dateModified: updatedAt ?? publishedAt`. A precedence inversion — both fields were present and
  valid, so every presence-based check passed.

Worth distinguishing in the output, because they point at different code: **ahead** means the field is
build-stamped, **behind** means stale or inverted precedence. Lastmod being older than the page's own
dateModified is the more clearly wrong of the two — it tells crawlers not to bother with a page that
has in fact changed.

## Acceptance Criteria
- [ ] New site-scope rule `crawl/sitemap-lastmod-drift` under `packages/rules/src/crawl/`, registered in the rules catalog
- [ ] For each crawled page present in the sitemap, compares sitemap `lastmod` against the page's own strongest date signal (schema `dateModified`, else `dateModified` equivalent meta, else visible date)
- [ ] Warns when `lastmod` is **older** than the page's `dateModified` by more than 7 days, reported as a distinct check name from the ahead case
- [ ] Warns when `lastmod` is **newer** than the page's `dateModified` by more than a configurable threshold (default 30 days)
- [ ] Each finding reports the two dates and the delta in days
- [ ] Skips pages with no page-side date signal at all (that is `content/freshness` territory, not drift)
- [ ] Does NOT warn on same-day or within-threshold differences
- [ ] Passes on a fixture where lastmod tracks dateModified per page
- [ ] Byte-identical output across the legacy and `siteQuery` paths
- [ ] Rule page added to `docs/` and `make validate` passes there
- [ ] `packages/audit-engine/tests/streaming-rules-canonical-golden.test.ts` updated: bump `v1.findings.length` and `v1.perRuleTally.length` by exactly one each, and add a comment line naming this rule as the cause, matching the convention already used for the six rules above it

nc9:
[...]

PR #112, opened 2026-08-19
feat(rules): add crawl/sitemap-lastmod-churn

### LINKED ISSUE #105: rules(crawl): sitemap lastmod churn — all values collapsed onto one build date
Part of #103 (V-01).

A sitemap whose `lastmod` values all land on one or two days is not telling anyone anything — the
field is being stamped at build time. Right now nothing in the catalog reads lastmod **values** at all
(`crawl/sitemap-valid` checks structure and only mentions lastmod in its solution prose), so this
ships silently.

**Found on two production sites, both real generator bugs, both invisible to the audit:**

- A 44-page personal site — all 44 sitemap URLs carried the build date.
- A 147-page service site — all 147 URLs carried the build date. After the fix: 11 distinct real dates.

**No new extraction needed.** `SitemapUrl.lastmod` is already parsed by the crawler and already
reaches rules via `site.sitemaps.discovered[].urls[]` (`packages/core-contracts/src/index.ts:1134`).

The judgement call is the guard. A 6-page brochure site legitimately published everything on one day,
and a site that emits no lastmod at all is a *different* finding — neither should warn here.

## Acceptance Criteria
- [ ] New site-scope rule `crawl/sitemap-lastmod-churn` under `packages/rules/src/crawl/`, registered in the rules catalog and exported from the rules index
- [ ] Warns when the count of distinct lastmod **days** across sitemap URLs is <= 2 and the sitemap carries at least 20 URLs with a lastmod
- [ ] The warn message states the distinct-day count and the URL count, and lists the offending dates
- [ ] Returns `skipped` (not `pass`) below the 20-URL floor, with a message saying the sitemap is too small to judge
- [ ] Returns `pass` on a fixture whose lastmod values spread across many days
- [ ] Does NOT warn on a fixture where sitemap URLs carry no `lastmod` element at all
- [ ] Does NOT warn when the sitemap has >= 20 URLs but only 3+ distinct days
- [ ] Runs off stored scalars so the legacy and `siteQuery` paths emit byte-identical checks
- [ ] Rule page added to `docs/` and `make validate` passes there
- [ ] `packages/audit-engine/tests/streaming-rules-canonical-golden.test.ts` updated: bump `v1.findings.length` and `v1.perRuleTally.length` by exactly one each, and add a comment line naming this rule as the cause, matching the convention already used for the six rules above it

nc9:
Closes #105

Adds `crawl/sitemap-lastmod-churn`, a site-scope rule that warns when sitemap `lastmod` values are collapsed onto a build date rather than reflecting real content changes.

## Acceptance Criteria
- [x] New site-scope rule `crawl/sitemap-lastmod-churn` under `packages/rules/src/crawl/`, registered in the rules catalog and exported from the rules index
- [x] Warns when the count of distinct lastmod days across sitemap URLs is <= 2 and the sitemap carries at least 20 URLs with a lastmod
- [x] The warn message states the distinct-day count and the URL count, and lists the offending dates
- [x] Returns `skipped` (not `pass`) below the 20-URL floor, with a message saying the sitemap is too small to judge
- [x] Returns `pass` on a fixture whose lastmod values spread across many days
- [x] Does NOT warn on a fixture where sitemap URLs carry no `lastmod` element at all
- [x] Does NOT warn when the sitemap has >= 20 URLs but only 3+ distinct days
- [x] Runs off stored scalars so the legacy and `siteQuery` paths emit byte-identical checks (rule reads only `ctx.site.sitemaps`, which is identical on both paths — no siteQuery branch needed)
- [x] Rule page added to `docs/` and `make validate` passes there

## Note
Sandbox in this session blocks execution of `bun`, so `bun test` / `bun x tsgo --noEmit` / `make validate` could not be run locally here. Implementation mirrors the existing `crawl/sitemap-domain` rule pattern exactly (same context shape, same skip/warn/pass conventions) and a matching test file (`packages/rules/tests/sitemap-lastmod-churn.test.ts`) was added following the `sitemap-coverage.test.ts` convention. Please confirm CI passes.

## Test plan
- [ ] `bun test packages/rules/tests/sitemap-lastmod-churn.test.ts`
- [ ] `bun x tsgo --noEmit` in `packages/rules`
- [ ] `make validate` in `docs/`

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -177,6 +177,7 @@
                   "rules/crawl/sitemap-4xx",
                   "rules/crawl/sitemap-coverage",
                   "rules/crawl/sitemap-domain",
+                  "rules/crawl/sitemap-lastmod-churn",
                   "rules/crawl/html-size",
                   "rules/crawl/pdf-size",
                   "rules/crawl/soft-404"
--- a/docs/rules/crawl/sitemap-lastmod-churn.mdx
+++ b/docs/rules/crawl/sitemap-lastmod-churn.mdx
@@ -1,0 +1,42 @@
+---
+title: "Sitemap Lastmod Churn"
+description: "Checks that sitemap lastmod values aren't all stamped with the same build date"
+---
+
+Checks that sitemap lastmod values aren't all stamped with the same build date
+
+| | |
+|---|---|
+| **Rule ID** | `crawl/sitemap-lastmod-churn` |
+| **Category** | [Crawlability](/rules/crawl) |
+| **Scope** | Site-wide |
+| **Severity** | warning |
+| **Weight** | 4/10 |
+
+## Solution
+
+If every URL in your sitemap carries the same (or nearly the same) lastmod date, the field is being stamped at build/deploy time rather than reflecting real content changes. This makes lastmod useless as a freshness signal - search engines may learn to ignore it entirely. Update lastmod only when a page's content actually changes, driven by your CMS or version control history, not by the deploy pipeline.
+
+## Enable / Disable
+
+### Disable this rule
+
+```toml squirrel.toml
+[rules]
+disable = ["crawl/sitemap-lastmod-churn"]
+```
+
+### Disable all Crawlability rules
+
+```toml squirrel.toml
+[rules]
+disable = ["crawl/*"]
+```
+
+### Enable only this rule
+
+```toml squirrel.toml
+[rules]
+enable = ["crawl/sitemap-lastmod-churn"]
+disable = ["*"]
+```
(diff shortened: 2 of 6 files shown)


nc9:
Pushed the missing piece rather than re-running the loop — an issue with an open PR must not be re-claimed, or a fresh session redoes finished work.

The rule itself was fine. CI blocked on `streaming-rules-canonical-golden.test.ts`, which pins both `v1.findings.length` and `v1.perRuleTally.length` so a new rule id can only land deliberately. Adding a site-scoped rule moves both by exactly one — 98216 → 98217 and 272 → 273 — and the convention is to record which rule caused the bump, which the six rules above it already do.

**This was a gap in the issue, not in the implementation.** The acceptance criteria never mentioned the golden tripwire, so the session had no way to know it existed. Every remaining rule-adding issue in the batch (#107, #108, #109, #110) now carries it explicitly, and I stopped the run before #107 could burn a full session into the same wall.

Verified locally: `bun test packages/audit-engine/tests/streaming-rules-canonical-golden.test.ts` → 1 pass.

PR #145, opened 2026-08-27
fix(rules): content/mojibake skips code, pre, samp and kbd

nc9:
## The false positive

`content/mojibake` read every character in `<body>`, so any page that *documents* encoding corruption failed the rule for doing so correctly.

Auditing squirrelscan.com warns twice, on `/releases` and `/releases/0.0.81`:

> `2 encoding corruption occurrence(s) in visible text (utf8-as-latin1): charset is correctly utf-8, so the source itself is double-encoded: re-import the content from its original source`

Both hits are the release note that announced the rule, which quotes the sequences it detects as examples:

```html
Catches garbled sequences like <code>â€™</code> and <code>Ã©</code>, and
replacement characters, in the text your visitors actually read.
```

Nothing on that page is corrupt. The rule was matching its own documentation, and its message sent the reader off to re-import content that is fine. Any changelog, docs page, or blog post about character encoding hits this.

## The fix

A code-like element is markup that says "this is a literal, not prose". `<code>` and `<pre>` are the common cases; `<samp>` (program output) and `<kbd>` (keys to press) make the same promise.

`packages/rules/src/content/text-content.ts` gains `getProseTextExcludingCode` next to the existing `getTextExcludingScripts`: the same non-mutating iterative walk (`collectTextExcluding` from `@squirrelscan/parser/extractors`), with `code`, `pre`, `samp` and `kbd` added to the excluded tags. Exclusion is by subtree, so a `<code>` nested mid-sentence does not leak into the scan and its prose siblings still do.

Only `content/mojibake` switches to the new helper. `keyword-stuffing`, `reading-level` and `hidden-text` share the old one and are deliberately untouched here, though the same argument may apply to some of them.

Detection in ordinary prose is unchanged, including prose on the same page as a quoted sample.

## Tests

New cases in `packages/rules/tests/mojibake.test.ts`:

- mojibake only inside `<code>` / only inside `<pre>` / inside `<samp>` and `<kbd>` passes
- `<p>text <code>Ã©</code> more</p>` passes: no leak from a nested span
- a `<code>â€™</code>` sample plus one real prose `â€™` warns with `value === 1`
- exclusion covers the whole subtree, not just direct text
- prose siblings of a code block are still read
- `U+FFFD` inside `<code>` does not escalate to a fail
- reading a code block does not mutate the shared DOM

`packages/rules/tests/content-no-dom-mutation.test.ts` gains the equivalence guard the old helper already had: `getProseTextExcludingCode` matches remove-then-`.textContent` for the code-like tags, and leaves the nodes in place.

The existing prose and clean-accents fixtures are unchanged and still pass.

## Verification

```
bun test                 4291 pass, 4 skip, 0 fail (358 files)
bun run test:engine      58 pass, 1 skip, 0 fail
bun run typecheck        clean, all 16 workspaces
bun run format:check     clean
bun run lint             clean
bun run docs:check       clean (tangly --strict)
bun run docs:build       clean, 387 pages
```

End to end against the live page, identical 70-page crawl:

```
# released 0.0.81 (before)
squirrel audit https://squirrelscan.com/releases --offline -m 70 -C full --rule-include content/mojibake
  66 passed - 2 warnings - 0 failed
  content/mojibake Encoding Corruption (warning)
    -> /releases
    -> /releases/0.0.81

# this branch (after)
bun run dev -- audit https://squirrelscan.com/releases --offline -m 70 -C full --rule-include content/mojibake
  68 passed - 0 warnings - 0 failed
```

Running the rule directly against the fetched `/releases` HTML: `utf8-as-latin1 count 2` before, `[]` after.

Docs updated: `docs/rules/content/mojibake.mdx` now lists both exclusion sets and why they differ.

### DIFF
--- a/docs/rules/content/mojibake.mdx
+++ b/docs/rules/content/mojibake.mdx
@@ -27,7 +27,12 @@ Three families of corruption, in visible text only:
 | `replacement-char` | `caf�` | the decoder gave up on those bytes |
 | `double-encoded-entity` | visible `&amp;nbsp;` | the source said `&amp;amp;nbsp;` |
 
-The check reads text with `<script>`, `<style>` and `<noscript>` excluded, never raw HTML, so an encoded sequence inside a script body is not reported. A page with no body is skipped rather than passed.
+The check reads prose, never raw HTML, so an encoded sequence in an attribute is not reported. Two kinds of markup are excluded:
+
+- `<script>`, `<style>` and `<noscript>`, because a visitor never reads them.
+- `<code>`, `<pre>`, `<samp>` and `<kbd>`, because a visitor reads them as a quoted literal. A changelog entry or docs page that quotes `â€™` inside a code span is documenting the corruption, not suffering from it.
+
+A page with no body is skipped rather than passed.
 
 **Severity depends on certainty.** The replacement character is unambiguous corruption and fails. The other families warn.
 
--- a/packages/parser/src/extractors/dom-text.ts
+++ b/packages/parser/src/extractors/dom-text.ts
@@ -16,10 +16,17 @@ const TEXT_NODE = 3;
  * removing the excluded elements then reading `.textContent` (comments and
  * processing instructions are never visited, matching `.textContent`), but
  * without mutating or cloning the DOM.
+ *
+ * `separator` is emitted in place of each skipped subtree. It defaults to "",
+ * which reproduces remove-then-`.textContent` exactly — including the fact that
+ * removing an element GLUES its neighbours together, so `Ã<code>x</code>©` reads
+ * back as `Ã©`. Callers that scan the result for character sequences must pass a
+ * separator, or they will see sequences that exist in neither fragment.
  */
 export function collectTextExcluding(
   root: Node,
-  isExcluded: (el: Element) => boolean
+  isExcluded: (el: Element) => boolean,
+  separator = ""
 ): string {
   const out: string[] = [];
   // Explicit stack of remaining child lists; index tracks position in each.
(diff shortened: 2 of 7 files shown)


nc9:
Review catch, fixed in c9a6638. It was a real regression introduced by the first commit.

`collectTextExcluding` joins surviving fragments with `""`, so skipping a subtree glues its neighbours:

```html
<p>Ã<code>literal</code>©</p>
```

extracted as `Ã©` and warned, on text that is clean in both fragments. Confirmed against the rule before the fix (`warn`, 1 occurrence) and after (`pass`).

**The boundary is a newline, not a space.** This matters more than it looks: `"Ã "` (à) and `"Â "` (non-breaking space) are themselves entries in `MOJIBAKE_SEQUENCES`, so a space boundary synthesizes a match against any preceding `Ã` or `Â` and just trades one false positive for another. Measured:

```
separator " "  -> findMojibake("Ã ") = [{kind:"utf8-as-latin1", sample:"Ã ", count:1}]
separator "\n" -> findMojibake("Ã\n") = []
```

No sequence in the table contains a newline. There is a regression test that fails if anyone later "simplifies" the boundary to a space.

**Existing callers are untouched.** `collectTextExcluding` gained an optional `separator` that defaults to `""`, and the push is guarded by `if (separator)`. `getCleanTextContent` and `getMainContent` stay byte-identical, which matters because `getCleanTextContent` feeds `contentHash`, and `getTextExcludingScripts` keeps its documented remove-then-`.textContent` equivalence for keyword-stuffing, reading-level and hidden-text.

Tests added:

- the glue reproduction above
- a fixture pinning the newline choice (`<p>Ã<code>sample</code></p>` and `<p>Â<pre>sample</pre></p>` must pass)
- `packages/parser/tests/dom-text.test.ts`, new, pinning the extractor contract: the default still glues and still matches remove-then-`.textContent`, a separator is emitted once per skipped subtree rather than per skipped text node, and the walk does not mutate the DOM
- the equivalence guard in `content-no-dom-mutation.test.ts` now asserts the honest invariant: ignoring inserted newlines, no surviving text is lost, gained, or reordered

Re-ran: `bun test` 4298 pass / 4 skip / 0 fail; `bun run test:engine` 58 pass / 1 skip / 0 fail; typecheck, lint and `format:check` clean. The live check still holds: `/releases` goes from 2 occurrences to 0.

One thing this does **not** fix, which predates the PR: `.textContent` semantics glue across *any* element boundary, so `<p><span>Ã</span><span>©</span></p>` warns on `main` today with no code element involved. Same root cause, wider blast radius, and out of scope here. Filed separately.

PR #188, opened 2026-09-01
feat(plans): free plan schedules one website weekly

nc9:
Public half of https://github.com/squirrelscan/repo/issues/1704 — the free plan gets a weekly scheduled audit on one website. The private API/dashboard half is a separate PR on the private monorepo and lands after this.

## What changes

Scheduling stops being the free/paid line. `scheduledCrawls` is now `true` on every plan, and two new `PlanDefinition` fields carry the real limits:

| | free | starter (Pro) / team / enterprise |
| --- | --- | --- |
| `maxScheduledWebsites` | `1` | `-1` (uncapped) |
| `scheduleFrequencies` | `["weekly", "monthly"]` | `["daily", "weekly", "monthly"]` |

Helpers: `planMaxScheduledWebsites`, `scheduledWebsiteCapReached`, `planAllowsScheduleFrequency`, `clampScheduleFrequency`, plus the `SCHEDULE_FREQUENCIES` ordering (most → least frequent) the clamp walks.

`clampScheduleFrequency` clamps DOWN to the nearest cadence the plan funds rather than rejecting — the same clamp-and-notify rule the per-audit page ceiling already uses. An unknown plan id resolves to free, i.e. the *tightest* limits, so a typo'd id can never buy more recurring spend.

## Why

A weekly audit spends roughly 250-300 of the free plan's 500 monthly credits, so the grant finally binds while the tier delivers recurring value. Daily is 4x that and would exhaust the grant in days, which is why it stays paid.

## Docs / copy

Every surface that sold scheduled audits as the Pro unlock was about to become false: the CLI pro pitch, the scheduled-audits page note, and the credits plan comparison now state the actual split. The "scheduling is off by default" line is also corrected — a new website starts weekly whenever the plan has a free slot, which is what the API has always done.

## Tests

`packages/core-contracts/tests/plans.test.ts` covers the per-plan limits, the frequency allowlists, the clamp direction (never up), the unknown-id fallback, and the cap boundary. `bun test` in `packages/core-contracts` and the full `apps/cli` suite (1889 pass) are green; `docs` passes `tangly check --strict`.

### DIFF
--- a/apps/cli/src/lib/upgrade.ts
+++ b/apps/cli/src/lib/upgrade.ts
@@ -49,7 +49,7 @@ export function proPitchLines(
 ): string[] {
   return [
     `  ${fmt.bold(PRO_HEADLINE)}`,
-    `  ${fmt.dim(`Also unlocks scheduled audits, faster crawls, and up to ${n(PRO.maxPagesPerAudit)} pages per audit.`)}`,
+    `  ${fmt.dim(`Also unlocks daily audits on every site, faster crawls, and up to ${n(PRO.maxPagesPerAudit)} pages per audit.`)}`,
     `  Upgrade: ${fmt.cyan(upgradeUrl(src))}`,
   ];
 }
--- a/docs/cli/credits.mdx
+++ b/docs/cli/credits.mdx
@@ -53,7 +53,7 @@ On the free plan the output ends with the Pro offer instead:
 
 ```text
   Pro: $19/month (or $190/year) for 3,000 credits a month
-  Also unlocks scheduled audits, custom request headers, and up to 2,000 pages per audit.
+  Also unlocks daily audits on every site, faster crawls, and up to 2,000 pages per audit.
   Upgrade: https://squirrelscan.com/upgrade?src=cli-credits
   Or run `squirrel credits --upgrade`.
 ```
--- a/docs/cloud/credits.mdx
+++ b/docs/cloud/credits.mdx
@@ -19,7 +19,7 @@ Annual billing is 12 months for the price of 10. Your monthly credits still arri
 
 Monthly credits are spent before purchased top-up credits. Subscribe and top up from the **Billing & Credits** page in the [dashboard](/dashboard).
 
-Free, Pro, and Team run the **same** audit at the same price: the base charge includes every cloud service (AI analysis, editor's summary, blocking check, tech + site metadata, domain stats, dead-link checking, report publishing) on every plan. Pro and Team just grant more monthly credits and unlock top-ups, [daily schedules on every website](/cloud/scheduled-audits) (free schedules one website, weekly), custom headers, and higher render concurrency - Team adds pooled credits, invites, and roles on top.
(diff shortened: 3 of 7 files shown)


nc9:
Added a third commit (d41b805) after review on the consuming private PR: `planScheduleSummary(planId)`, the one derivation of "what does this plan schedule" ("Weekly, 1 site" / "Daily, all sites").

Two pricing tables in two different private apps render that cell, and both derived it from the plan's own limits so a change in `PLANS` could not leave them lying. Each had its own byte-identical copy of the formatter, which quietly gave up the guarantee the comments on both were making: copies drift. It belongs next to `PLANS` for the same reason `planAtLeast` and `isSelfServePlan` do.

nc9:
Acceptance review addressed: bd0629b (the shape) and 8674d3a (the docs).

**1. MEDIUM — version skew on the wire shape. Fixed, and it was a real hazard rather than a theoretical one.** `PlanDefinition` is not only a `PLANS` entry; it is also `CreditsResponse.plan`, which cloud-client returns as a plain-JSON cast. That cast is additive-safe and **not** subtractive-safe, so a CLI built against these fields talking to a server that predates them (or has been rolled back) held a type promising values absent at runtime, and the first schedule check was `undefined.includes(...)` — a crash, not a wrong answer.

Both fields are now optional, behind one reader:

```ts
resolvePlanScheduleLimits(plan) // → { maxScheduledWebsites, scheduleFrequencies }
```

The fallback is the **free tier's** limits (1 site, weekly/monthly), never a permissive one: a plan we cannot fully read must degrade to the least entitlement rather than quietly granting daily audits on every site. Every `planId` helper (`planMaxScheduledWebsites`, `planAllowsScheduleFrequency`, `clampScheduleFrequency`, `planScheduleSummary`) routes through it, plus a new `planScheduleFrequencies`, so the safety is uniform instead of per-call-site. Tests cover: both fields absent, each absent independently, no plan at all, a complete plan passing through untouched, and an explicit `maxScheduledWebsites: 0` not being mistaken for an absent field (0 is falsy — that one would have silently handed a no-schedule plan the free tier's slot).

**Ripple into the private PR: three call sites, all mechanical**, pushed as squirrelscan/repo@2868b6a0 — the scheduler's cadence gate, the schedule settings page's allowed-cadence list, and the MCP upgrade payload. Nothing else read the fields off an object; everything else already went by plan id. `grep` for `.scheduleFrequencies|.maxScheduledWebsites` across `apps/*/src` now returns only reads off the resolver's result.

**2. LOW — `planScheduleSummary` labelling. Fixed.** It tested for `daily` and fell through to `"Weekly"`, so a monthly-only plan (valid under the type) would have been advertised at a cadence it does not run. It now takes the most frequent cadence the plan actually funds off the ordered `SCHEDULE_FREQUENCIES` list, and returns `"Not included"` — plain words, no em dash, per the public-copy rule — when the plan cannot schedule, has no cadences, or has zero slots.

**3. LOW — contradicting docs. Both fixed.** `docs/cli/credits.mdx` quoted a Pro pitch the binary no longer prints; its sample output now matches `apps/cli/src/lib/upgrade.ts` verbatim again. `docs/cloud/credits.mdx` listed scheduling among Pro's unlocks and gave free's cadence as weekly only; it now leads with scheduling being on every plan and names both free cadences. `docs/cloud/scheduled-audits.mdx` already stated both correctly.

Green locally: core-contracts 28 tests, `apps/cli` 1889, typecheck clean in core-contracts / cloud-client / cli, `tangly check --strict` passes. Private side: api 3413, dashboard 203, web 372, all typechecks clean against this branch.

PR #191, opened 2026-09-02
fix(audit-engine): keep the fetch deadline armed through the body read

nc9:
Five fetchers cleared their `AbortController` timer where the fetch resolved, i.e. when the response **headers** landed, leaving the body read with no time bound at all. `readBodyCapped` and `response.text()` cap **bytes**, not seconds, so an origin that answers 200 and then trickles or stalls its body parked the caller forever.

Same bug class as #187, which fixed it in the crawler. These are the stragglers.

## Sites fixed

| file | unbounded read |
|---|---|
| `packages/audit-engine/src/script-fetcher.ts` | `readBodyCapped` after `clearTimeout` |
| `packages/audit-engine/src/external-checker.ts` | `getResponse.text()` on a 403, for WAF detection |
| `apps/cli/src/crawler/script-fetcher.ts` | `response.text()` (fork of the engine fetcher) |
| `apps/cli/src/crawler/external-checker.ts` | same WAF read (fork) |
| `apps/cli/src/crawl/sitemaps.ts` | `response.text()` in a later Effect stage, outside the timer's scope |

The two CLI files are **forks**, not re-exports: `apps/cli/src/audit/adapter.ts` imports `@/crawler/*`, so they are what the `squirrel` binary actually runs. Fixing only the package would have left every CLI user hanging.

## Fix

`clearTimeout` moves into a `finally` that wraps the body read, matching `packages/crawler/src/deadline.ts` and `packages/fetchers/src/index.ts`. The script fetchers additionally split their retry loop out of the attempt, so each retry arms a fresh deadline rather than inheriting a spent one or keeping it alive across the backoff.

## Impact

Scripts and external links are fetched during rules enrichment, **after** pages are collected, so the failure mode was the post-crawl hang class rather than a crawl-phase wedge: the audit has all its pages and never finishes. The sitemap read hangs the crawl preamble instead.

Behaviour changes only on the previously-hanging path. A stalled script body now returns `error: "timeout"`; a stalled WAF 403 body returns the plain 403 it already knew from the headers (the WAF verdict degrades to unknown, matching the existing "ignore body read errors" branch); a stalled sitemap returns a failure result.

## Tests

`apps/cli/tests/crawler/fetch-deadline.test.ts` and `packages/audit-engine/tests/fetch-deadline.test.ts` serve headers immediately, then hold the body stream open forever. Verified by reverting each source file to `main` and re-running: all five cases hang and fail on Bun's test timeout without the fix, and return within the deadline with it. Each fetcher also has a fast-path test proving normal responses are still read in full.

`packages/audit-engine/package.json`'s `test:golden` glob gains `tests/*deadline*.test.ts` — the CI engine job runs only that script, so a new test file would otherwise never execute in CI.

- `bun test` in `apps/cli`: 1903 pass, 0 fail
- `bun run test:engine`: 66 pass, 0 fail, golden baseline unchanged (`pages=518 findings=98221 healthScore=48`)
- `bun run typecheck` (all workspaces), `lint`, `format:check`: clean

No HTML output bytes change, so `REPORT_HTML_VERSION` is untouched.

## Follow-up, not in this PR

`apps/cli/src/tools/request.ts` (`request` / `requestOnceWithTiming`) has the same shape: it clears the deadline and hands the `Response` to callers that read the body later (`robots.ts`, reachability, and others). Fixing it properly means moving those call sites to a `withRequestDeadline(timeoutMs, perform, use)` contract, which is a wider refactor than this PR should carry. Filed separately.

Refs squirrelscan/repo#1729

### DIFF
--- a/apps/cli/src/crawl/sitemaps.ts
+++ b/apps/cli/src/crawl/sitemaps.ts
@@ -136,6 +136,9 @@ export type SitemapFetchResult =
   | { success: true; data: SitemapData }
   | { success: false; url: string; error: string };
 
+/** Deadline for one sitemap fetch, covering the body read (#1729). */
+export const SITEMAP_FETCH_TIMEOUT_MS = 30000;
+
 /**
  * Fetch a single sitemap
  * Uses standard fetch (not IMPIT) since sitemaps are meant to be crawled by bots
@@ -143,13 +146,18 @@ export type SitemapFetchResult =
  */
 export function fetchSitemap(
   url: string,
-  userAgent: string
+  userAgent: string,
+  timeoutMs: number = SITEMAP_FETCH_TIMEOUT_MS
 ): Effect.Effect<SitemapFetchResult, never, never> {
   return pipe(
     Effect.tryPromise({
+      // #1729: the deadline covers the BODY read, not just the headers. The
+      // text read used to sit outside the timer's scope, so a sitemap host that
+      // answered 200 and then trickled or stalled its body had no time bound at
+      // all and parked the crawl preamble.
       try: async () => {
         const controller = new AbortController();
-        const timeoutId = setTimeout(() => controller.abort(), 30000);
+        const timeoutId = setTimeout(() => controller.abort(), timeoutMs);
 
         try {
           const response = await fetch(url, {
@@ -161,7 +169,13 @@ export function fetchSitemap(
             signal: controller.signal,
             redirect: "follow",
           });
-          return response;
+          if (!response.ok) {
+            // Nothing here reads a non-2xx body; drop it so the connection is
+            // released rather than left half-consumed.
+            void response.body?.cancel();
+            return { ok: false as const, status: response.status };
+          }
+          return { ok: true as const, content: await response.text() };
         } finally {
           clearTimeout(timeoutId);
         }
@@ -172,41 +186,14 @@ export function fetchSitemap(
         error: error instanceof Error ? error.message : "Network error",
       }),
     }),
-    Effect.flatMap((response) => {
-      // Network error or non-2xx status
-      if (!response.ok) {
-        const errorMsg =
-          "status" in response && response.status
-            ? `HTTP ${response.status}`
-            : "error" in response
-              ? String(response.error)
-              : "Network error";
-        return Effect.succeed<SitemapFetchResult>({
-          success: false,
-          url,
-          error: errorMsg,
-        });
+    Effect.map((result): SitemapFetchResult => {
+      if (!result.ok) {
+        return { success: false, url, error: `HTTP ${result.status}` };
       }
-
-      return Effect.tryPromise({
-        try: () => response.text(),
-        catch: () => null,
-      }).pipe(
-        Effect.map((content): SitemapFetchResult => {
-          if (!content) {
-            return {
-              success: false,
-              url,
-              error: "Empty response",
-            };
-          }
-          const parsed = parseSitemap(content, url);
-          return {
-            success: true,
-            data: parsed,
-          } as SitemapFetchResult;
-        })
-      );
+      if (!result.content) {
+        return { success: false, url, error: "Empty response" };
+      }
+      return { success: true, data: parseSitemap(result.content, url) };
     }),
     Effect.catchAll(() =>
       Effect.succeed({
--- a/apps/cli/src/crawl/sitemaps.ts
+++ b/apps/cli/src/crawl/sitemaps.ts
@@ -182,11 +182,19 @@ export function fetchSitemap(
           clearTimeout(timeoutId);
         }
       },
-      catch: (error) => ({
-        status: 0,
-        ok: false,
-        error: error instanceof Error ? error.message : "Network error",
-      }),
+      // Deliberately NOT shaped like the success branch above: this value goes
+      // to the ERROR channel, so it never reaches the mapper. Giving it an
+      // `ok`/`status` of its own invited the misreading that a failed fetch
+      // could surface as "HTTP 0".
+      catch: (error) => {
+        const err = error as Error;
+        // #1729: the deadline now covers the body read too, so name that case
+        // instead of letting a stalled sitemap read as a generic failure.
+        if (err?.name === "AbortError") {
+          return { reason: `Timed out after ${timeoutMs}ms` };
+        }
+        return { reason: err?.message || "Network error" };
+      },
     }),
     Effect.map((result): SitemapFetchResult => {
       if (!result.ok) {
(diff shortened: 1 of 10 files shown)


nc9:
Acceptance findings addressed in 37ac3b3. Two of the four needed a correction to the finding itself; details below.

## HIGH — `packages/crawler/src/fetcher.ts` — real, fixed

Confirmed and fixed. `requestWithTiming` had the same clear-at-headers bug wearing a `try/finally` that looks correct: the `finally` closes at the fetch, but the `Response` escapes and its body is read much later in `fetchPageStandard`. The same `finally` also removed the fiber-interrupt bridge, so an interrupt could not reach the reader.

**One correction to the framing:** it was not an unbounded hang. `crawler.ts:1299` wraps `processUrl` in `Effect.timeout(max(120_000, timeoutMs * 6))`, so a stalled page was bounded by the per-URL watchdog. The damage was that a stall burned the whole watchdog (at least 120s) instead of `timeoutMs`, and that the watchdog's abort never cancelled the read: `readBodyCapped` stayed parked on `reader.read()` for the life of the process, holding the socket and its buffered chunks.

`requestWithTiming` now returns a `TimedResponse` whose deadline stays armed. The caller owns it across the redirect loop and the body read, drops superseded hops (which the loop never did, leaking their sockets), and `Effect.ensuring` releases whatever is still held on every exit — a `try/finally` inside `Effect.gen` does not run on failure. The body read's interrupt bridge now aborts the request instead of only losing a race, and a deadline firing mid-body reports a timeout rather than a parse error. `fetchPageStandard` is split behind `Effect.suspend` so the held state is per **run**, since callers retry this same Effect value.

Both new cases fail against `main`: the stall case by hanging, and the interrupt case because the test server observes that the client never aborted (`req.signal`).

Not touched: `deadline.ts`, per your note about #194.

## MED — WAF 403 — real, but the first fix I wrote was inert

Your diagnosis was right, and my initial fix was wrong in a way worth recording. I added a `wafCheckInconclusive` flag honored by `filterBrokenLinks`. A codex pass caught that **`filterBrokenLinks` is not the production path**: results flow through the adapters into `storage.upsertLink` (which has no such column) and are re-filtered by `packages/rules/src/links/broken-external-links.ts`. The flag would have been dead code that looked like a fix. Plumbing it properly needs a new SQLite column on `links` plus changes across 7 sites in 3 packages — too much for this PR, and a schema change I would not slip into a deadline fix.

What landed instead is smaller and actually reaches production: on a failed 403 body read, **run WAF detection on the response headers** rather than abandoning it. `detectWaf(headers)` treats an absent body as headers-only, and most WAFs are identifiable from `cf-ray` / `x-sucuri-id` / a provider `server` value alone. That is real evidence, not a guess, and it needs no new field. A 403 that stalls *and* carries no WAF header still reads as a plain 403 — the pre-existing ambiguity, which predates this PR (any body-read failure already fell through that way) and is now the only remaining gap.

## MED — sitemap reason — substance right, mechanism wrong

`HTTP 0` is not reachable. `Effect.tryPromise`'s `catch` value goes to the **error channel**, so it never reaches `Effect.map`; `result.status` in the mapper is always a real HTTP status. Verified by running it — a stalled body and a DNS failure both produced `"Unexpected error"` (from the terminal `catchAll`), and a 404 produced `"HTTP 404"`.

The substance holds: the reason *was* discarded, just at `catchAll` rather than at the mapper. `catchAll` now carries it, and the error-channel value no longer mimics the success shape — that overlap is what made `HTTP 0` look reachable. The three modes are now distinct: `Timed out after 300ms`, `HTTP 404`, and the runtime's connect message.

## LOW — retry freshness — added, and the first version was too weak

Added to both suites. Codex flagged that my first version passed against unfixed code, which was correct: the old recursive implementation already built a fresh controller per attempt. Attempt 2 now hits a **stalled** body, which pins three things at once — a retry happened, the retry's deadline covers its body read, and that deadline is fresh (a spent timer or aborted controller fails attempt 2 instantly, at about the backoff rather than backoff plus deadline). Both now fail against `main`.

## Verification

Every new case was run against `origin/main` sources, not just asserted: 4 of 5 engine cases and 6 of 8 CLI cases fail there, plus both crawler stall cases.

- `packages/crawler`: 268 pass
- `packages/audit-engine`: 317 pass, 3 skip
- `apps/cli`: 1906 pass
- `bun run test:engine`: 68 pass, golden baseline unchanged (`pages=518 findings=98221 healthScore=48`)
- root `typecheck`, `lint`, `format:check`: clean

nc9:
Both round-2 findings confirmed and fixed in cea5229. Branch rebased onto main after #193 merged.

I reproduced each with an instrumented probe **before** touching the code, so these are measured rather than reasoned:

```
A terminal 403      -> type: blocked | server saw cancel: false
B failed hop        -> type: network | server saw cancel: false
C stalled 503       -> type: network | msg: Server error: 503
```

and after:

```
A terminal 403      -> type: blocked | server saw cancel: true
B failed hop        -> type: network | server saw cancel: true
C stalled 503       -> type: timeout | msg: Crawl request timed out
```

## HIGH — `release()` left unread bodies parked

Correct on both counts. `release()` only disarmed the timer, so every path that releases *without* a body read leaked the socket:

- a terminal 403/429/5xx fails `applyStatusGuards` before `readBody` is ever yielded;
- a redirect hop is still held when the **next** request fails, because the loop drops a superseded hop only once a newer response arrives.

`release()` now aborts as well as disarms. That is safe unconditionally rather than conditionally, and the reason is worth stating: `release` runs only from `Effect.ensuring`, i.e. as the effect leaves, after the gen body has already produced its return value and extracted everything it needs from the response. So there is no path where a body is read *after* release. On the success path the stream is already drained and the abort is a no-op; on every failure path nobody will ever consume it. No consumed-flag bookkeeping needed, and none added.

## MED — 503 challenge read swallowed the timeout

Correct. `Effect.orElseSucceed(readBody, () => undefined)` caught everything, including the timeout this PR newly maps, so a stalled 503 surfaced as `network: Server error: 503` — the origin blamed for our own deadline. It now re-fails on `error.type === "timeout"` and still degrades every other read failure to "no body", so the guard can classify the 503 from `cf-mitigated` and friends alone.

## Tests

Three added to `packages/crawler/tests/fetch-body-deadline.test.ts`, all run locally and **each verified to fail against the previous commit (37ac3b3)**:

- `a terminal status released without a read still cancels the connection` — server-side `req.signal` observer, 403
- `a hop still held when the next request fails cancels that hop` — 302 to a closed port, same observer
- `a stalled 503 body is classified as a timeout, not a server error`

The two cancellation tests assert what the server *observed*, not what the client returned, since the client returns promptly either way — that distinction is what made the original leak invisible.

## Verification

- `packages/crawler`: 271 pass (was 268; +3)
- `apps/cli`: 1907 pass
- `bun run test:engine`: 68 pass, golden baseline unchanged (`pages=518 findings=98221 healthScore=48`)
- root `typecheck`, `lint`, `format:check`: clean

Will confirm from the CI log that the three new cases executed there, not just locally.

PR #196, opened 2026-09-03
fix(crawler): crawl-wide rate-limit backoff; stop reporting throttled URLs as broken (#1829)

nc9:
Fixes the false "broken 4xx" reports a multi-storefront operator was seeing: those pages and links were rate limited, not dead. Tracked privately as squirrelscan/repo#1829.

## What changed

**Classification.** 429, Shopify's 430, and a 503 carrying `Retry-After` are all rate limiting. A bare 503 stays a server error, and a Cloudflare challenge 503 stays `blocked` (a challenge never clears on retry, so backing off would waste the whole budget on a wall).

**Backoff is crawl-wide, not per request.** Previously a 429 retried 3 times over ~1.5s while the other four workers kept firing at the same host, so one refusal cascaded. Now a single rate-limit response pauses every worker aimed at that host, drops it to one in-flight request, and multiplies its politeness delay. `Retry-After` is honoured as the minimum wait in both its seconds and HTTP-date forms; without it the first wait is 5s and doubles with jitter for up to six attempts. Recovery is earned back in steps over 20 consecutive 2xx. State is per host, so external link checks elsewhere are untouched.

**New `[crawler] max_backoff_ms`, default 300000.** Caps a single wait and the cumulative wait for one URL and one host. Past it, the host's remaining URLs are recorded as rate limited without further requests and the crawl finishes.

**Throttled URLs are never "broken".** `links/broken-links`, `links/broken-external-links`, `links/dead-links`, `links/redirects` and `crawl/sitemap-4xx` move them into an info bucket reading "status unverifiable". Rate-limited pages are excluded from page-level scoring the way WAF challenge pages already are. Crawl stats carry `pagesRateLimited` separately from `pagesBlocked` and `pagesFailed`; 429 is no longer folded into `pagesBlocked`.

**The report says whether its numbers cover the site.** A crawl that gathered content but lost pages reports `status: partial` with a reason naming the count and host, and keeps its score. A crawl whose entry URL was only ever throttled reports `blocked` with a rate-limit reason rather than the bot-protection one, which points at the opposite fix. A machine-readable `rateLimited: { pages, hosts }` rides along, and the text, markdown, llm, json, html and CLI console renderers all print it.

**Progress visibility.** A new `rate-limited` crawler event plus a synchronous `onRateLimitEvent` hook, so the CLI progress line says which host is throttling and for how long while it waits, instead of a counter that silently stops.

**Docs.** `max_backoff_ms` reference plus a "Crawling rate-limited hosts" guide with the Shopify recipe. `make validate` passes.

## End-to-end validation

Real audits from source (`bun run src/cli.ts audit …`), `--offline --refresh`, against public Shopify storefronts. Numbers below are measured, not estimated.

### Rate limiting provoked (colourpop.com, 20 concurrency / 0ms delay / 40 pages)

| | before (origin/main) | after |
|---|---|---|
| wall time | 3.3s | 61.8s |
| pages | 0 | 0 |
| status | `blocked` | `blocked` |
| statusReason | `Site blocked the crawler (bot protection / auth / rate limit)` | `Rate limited by colourpop.com; no pages could be fetched` |
| `rateLimited` | absent | `{"pages": 1, "hosts": ["colourpop.com"]}` |
| exit code | 0 | 0 |

The before run gives up in 3.3s and blames bot protection, sending the reader to their WAF. The after run rides out the backoff schedule and names throttling.

CLI output from the after run:

```
[+0.2s] warn Site uses cloudflare protection - some pages may be inaccessible and results may be incomplete
New crawl: https://colourpop.com
[+1m2s] warn rate limited colourpop.com throttled this crawl; set [crawler] per_host_concurrency = 1 and per_host_delay_ms = 500 in squirrel.toml to crawl it politely
✓ Audited 0 pages in 61.8s
```

### Mid-crawl throttling with the default 5-minute cap (www.allbirds.com, 20 concurrency / 0ms delay / 100 pages)

```
[+2m35s] warn rate limited www.allbirds.com throttled this crawl; set [crawler] per_host_concurrency = 1 and per_host_delay_ms = 500 in squirrel.toml to crawl it politely
✓ Audited 93 pages in 189.1s
```

```
status       : partial
statusReason : 907 pages rate limited by www.allbirds.com
pages        : 93
score        : 41 (kept, not nulled)
broken-link / dead-link / sitemap-4xx findings from the throttling: 0
```

That run is what surfaced the `907` bug: a rate-limited record stored no page, so the maxPages budget never filled and the dispatch loop walked the entire 1000-URL sitemap marking rows. Rate-limited records now count against `max_pages`, so the figure reflects the budget the crawl actually had. A regression test pins it.

The same target with `max_backoff_ms = 45000` returned 0 pages: the shorter cap gives up before the limiter relents. That is the setting working as designed, and it is why the default is five minutes.

### No regression on healthy sites

| | before | after |
|---|---|---|
| www.gymshark.com, 40 pages | `completed`, 40 pages, score 39, 86 issues | `completed`, 40 pages, score 39, 86 issues |

Also clean after the change: gymshark at 150 pages (206s) and 250 pages (314s), both `completed` with no `rateLimited` block; allbirds at 40 pages, `completed`, score 41.

### Honest gaps

- I could not provoke a **mid-crawl** throttle on colourpop or gymshark. When colourpop's limiter engages it refuses the seed outright, so those runs land in `blocked` rather than `partial`. The `partial` path is covered by the allbirds run above plus unit tests.
- `Retry-After` was never sent by any of the three storefronts during these runs, so the header path is covered by tests (seconds, HTTP-date, past date, clamping) rather than by live traffic.
- Under an aggressive profile the crawl preamble alone (robots.txt, sitemaps, the agent-file probes) can trip a limiter before page 1, which is why the throttled runs report 0 pages rather than a partial.

## Review round

A codex pass found ten real issues, all fixed in the third commit, each with a regression test that fails without the fix:

- Two workers refused at the same moment each extended the host window, but each slept only its own increment, so the first retried while the second still waited. `noteRateLimited` now returns the host's full remaining window.
- A worker could pass the pre-dispatch wait, block in `acquire()`, and start without re-checking. Re-checked after acquire, inside the guaranteed-release block, bounded by the per-URL watchdog.
- `Retry-After: 0` (or a past HTTP-date) produced a hot retry burst that never accumulated toward the cap, so the host could never be given up on. Every wait is now floored at 1s.
- The attempt cap was checked before notifying the registry, so the final refusal was never reported and strike counts under-counted.
- `onSuccess` fired for every successful fetch, so twenty 404s walked a throttled host back up. Only 2xx counts now.
- The registry was created once per crawler instance and never reset; since an exhausted host is skipped before any request it can never recover, so a second crawl recorded every URL untried.
- Resume zeroed the rate-limited budget, letting a resumed crawl re-spend it. Seeded from persisted stats.
- Throttled link results were cached for the full 7-day TTL, pinning a live link as "unverifiable" for a week. They are no longer cached, and the column added for them is dropped rather than left as a field nothing writes.
- The smart-audit status override replaced a rate-limit `partial` with `completed`.
- `maxBackoffMs` was added to the persisted snapshot type but never written; NaN options could poison the scheduler.

## Follow-up needed in the private repo

`DeadLinkResult` (`packages/core-contracts/src/services.ts`) gains an optional `rateLimited` flag, and the CLI derives it from 429/430 when absent. The **cloud dead-links service must populate it** for the `503` + `Retry-After` case, which cannot be recovered from the status code alone. Until then a throttled 503 seen by the cloud bulk checker still reads as an outage. I will open that PR on the private repo separately.

## Tests

New: rate-limit classification and `Retry-After` parsing, the per-host registry on a fake clock (crawl-wide pause, concurrency drop, recovery sequence, cumulative cap, exhaustion, host isolation, reset, NaN options), the fetcher retry schedule under Effect's `TestClock`, a crawl against a host that starts throttling, the five link/sitemap rules, page-level scoring exclusion with a 200-status control, config validation, and the partial/blocked status reasons.

Regressions run green: crawler (status guards, full-crawl harness, slow origin, preamble budget, deadline, stop hard cap, zero-success breaker, schema migrations), audit-engine (golden baseline, v2 report golden, streaming page and collector rules, site-query link rules, redirect-chain e2e, scoring), rules (blocked links, sitemap valid/coverage, anchor text), report (failure notice, html/json/llm status) and CLI (llm output, scoring, link analysis, config).

Note for reviewers: `packages/audit-engine/tests/report-stream-v2-golden.test.ts` gets killed by a local memory watchdog when its three tests run in one process (each rebuilds a 518-page fixture). Each passes individually in ~8s. That is a pre-existing cost of the fixture, not something this branch introduces.

Closes #1829


nc9:
**`Dependency audit` failure is unrelated to this branch.**

Six newly published advisories in transitive dependencies:

- `qs` — GHSA-x5fp-wj9c-mxmx, GHSA-4mjr-xmp4-gh2g (moderate; via express/body-parser)
- `fast-uri` — GHSA-5jgf-p345-68v8, GHSA-f65p-4m7j-42xc, GHSA-fph4-wmhf-6fwf, GHSA-jqff-g426-hqxp (high; via ajv)

Evidence this branch did not introduce them:

- `git diff origin/main -- bun.lock` is **empty**. The only package.json change in the branch adds a subpath export (`@squirrelscan/utils/rate-limit`), no dependency.
- `bun run audit:dependencies` reproduces the identical six advisories locally, and since the lockfile is byte-identical to `main` it fails the same way there.
- `main`'s last CI run (3f7b57d) predates these advisories, which is why it was green.

Fixing it means either bumping `ajv`/`express` or extending the `--ignore` allowlist in `audit:dependencies`. Both are maintainer calls about the dependency policy rather than part of this change, so I have deliberately not silenced them here. Happy to do it in a separate PR if you want the gate green first.

nc9:
Cloud-side follow-up is open: **squirrelscan/repo#1834** (private).

It declares `rateLimited` on the API publish schema and per sitemap URL status, and shows the count on the dashboard report page. Worth flagging because the API schema is a plain `z.object` and **zod strips undeclared keys** — without that PR both new fields are dropped silently on publish and the CLI-side work here is inert for cloud audits.

Two items remain after both land, since each needs the types from this PR plus a gitlink bump:

1. `packages/cloud-audit-runtime` — forward `max_backoff_ms` into the crawler config and expose it per website. The 5-minute default applies today; a per-website value cannot be chosen yet.
2. The cloud dead-links service — set `rateLimited` for the `503` + `Retry-After` case. This PR derives it from 429/430 as a fallback, so only the 503 case is currently lost on the cloud bulk path.

Verified that `packages/cloud-audit-runtime` and `apps/worker-agent` typecheck against this branch, so nothing on the private side is forced by merging it.

nc9:
Two notes on how the acceptance criteria map onto this repo, so they can be checked without guessing:

**`cwv` is not a rule id.** The criteria name `soft-404`, `indexability`, `cwv` and `http-to-https` as page rules that must produce no finding for a 429 page. The real ids are `crawl/soft-404`, `crawl/indexability` and `security/http-to-https`; `packages/rules/src/performance/cwv.ts` is a shared helper (`getCWVHints`) with no rule of its own, feeding `perf/cls-hints`, `perf/inp-hints`, `perf/preconnect` and `perf/font-loading`.

That naming does not weaken the guarantee, because the exclusion is not per rule. A rate-limited page is dropped from `parsedPages` in the adapter, which is the corpus **every** page-scope rule runs over, so no page rule can see it. The test in `packages/audit-engine/tests/rate-limited-pages-e2e.test.ts` asserts on the keys of `pageResults` (the set of pages actually graded) rather than on any one rule's output, with a control that proves the same body IS graded and indicted at status 200.

**Synthetic 429 site.** The criterion asks for a smoke run against a synthetic site that returns 429 for a burst then recovers. I covered that as a crawl test rather than a `packages/synthetic-site` fixture: `rate-limit-crawl.test.ts` drives a real crawl through `config.documentFetcher` whose first response is 429 and the rest 200, asserting a complete crawl, zero rate-limit losses and exactly one backoff. Same shape, no fixture-generator dependency. Happy to move it into synthetic-site if you would rather it lived there.

nc9:
Precision on the e2e evidence, so nothing in the description reads as stronger than it is.

**The `partial` run showing 93 pages / "907 pages rate limited" is from BEFORE the budget fix.** It is the run that surfaced the bug, not a demonstration of current behaviour. After the fix I could not re-observe a live `partial`: by then both allbirds and colourpop were refusing the seed outright, so re-runs landed in `blocked` with `rateLimited: {pages: 1}` — correctly bounded, but not the mid-crawl shape.

So the post-fix bound on that count is pinned by a regression test rather than by live traffic:

> `recorded rate-limited URLs count against max_pages, so the frontier is not drained` — a 60-URL site with `maxPages: 5`, asserting `pagesFetched + pagesRateLimited <= 5`.

Full run matrix, all measured:

| run | status | pages | score | rateLimited.pages |
|---|---|---|---|---|
| before gymshark 40p | completed | 40 | 39 | 0 |
| after gymshark 40p | completed | 40 | 39 | 0 |
| after gymshark 150p | completed | 150 | 40 | 0 |
| after gymshark 250p | completed | 250 | 40 | 0 |
| after allbirds 40p | completed | 40 | 41 | 0 |
| before colourpop 40p | blocked | 0 | N/A | absent |
| after colourpop 40p | blocked | 0 | N/A | 1 |
| after allbirds 100p (pre-fix, default cap) | partial | 93 | 41 | 907 |
| after allbirds 100p (post-fix) | blocked | 0 | N/A | 1 |

The two `before` rows are the only true A/B pairs — a third-party limiter's state moves between runs, so the allbirds before/after is not comparable and I have not presented it as one.

All runs used `--offline`, so no credits were spent and nothing was published.

PR #197, opened 2026-09-03
chore(cli): pause domain stats

nc9:
Pause the domain-level SEO summary (backlinks, referring domains, organic keywords and traffic, domain rank). It was the one part of a report that described a site's market position rather than anything you could fix in the code, and it sat awkwardly next to everything else squirrelscan tells you.

Paused, not removed. `cloud.domain_stats` stays in the config and now defaults to `false`. The provider call, the report renderers, the contracts and the types are all untouched, so bringing it back is the default flip.

## What changes for users

- An audit runs no domain stats step, and reports carry no domain stats section.
- The price is unchanged. The flat 50-credit base already covered it, so there is nothing to refund and nothing new to pay for.
- Audits that already captured domain stats keep the section in their published reports.
- Docs and the `squirrel credits` help text stop listing it among what the base includes. The help string and both docs samples are byte-identical, as they were before.

## The case that needed care

A config written before this change still says `domain_stats = true`. That user's audit reaches a hosted route that now answers 404, and `runCloudDomainStats` degrades to `null` like any other failure: no section, no spend, no throw. But it also printed two progress lines on every audit:

```text
cloud: domain stats
cloud: domain stats skipped
```

A step that can only 404 should not narrate itself, so those are gone. The failure is logged at debug level instead. `onProgress` stays on the options interface with a note saying to restore the pair if the service comes back.

## Tests

- `apps/cli/tests/config.test.ts` pins the default and that an explicit `true` is still honoured.
- `apps/cli/tests/audit/domain-stats-paused.test.ts` drives the real `runCloudDomainStats` and watches the client, the spend callback and the progress stream: default → no call at all; explicit opt-in against a paused 404 → one call, then silence; and a live service still returns stats, so the pause is a default rather than a dead branch.
- `apps/cli/tests/reports/output/console-domain-stats-paused.test.ts` checks the console report shows no heading and none of the section's vocabulary, and still renders the section for a report that has the data.

## Paired change

The hosted service is switched off separately behind a server-side flag. Deploy order does not matter: the CLI skips the step by default, and an opted-in config degrades quietly either way.


nc9:
**Dependency audit is red for a reason unrelated to this PR, and it is red on every open PR.**

`bun audit` picked up six advisories published since the last green run on main (2026-09-02 09:38 UTC): four high in `fast-uri` (host confusion and SSRF via IDN/IPv6/percent-decoding handling) and two moderate in `qs`.

This branch changes no dependency at all:

```
git diff origin/main...HEAD -- package.json bun.lock 'packages/*/package.json' 'apps/*/package.json'
```

returns empty, so the lockfile the audit resolves against is byte-identical to main's. PR #196, an unrelated crawler change, fails the same check.

Both transitive deps need a bump (or an ignore entry) on main. That is a separate call and a security one, so I have not folded it into this PR.

PR #224, opened 2026-09-07
fix(audit-engine): bound the complete-store finalize to rules, not findings

nc9:
## What

The chunked publish path scores off the COMPLETE per-page findings a cloud audit streams into the finding store. It got there by materializing them: every row resident at once, reconstructed into per-page `CheckResult`s, unioned, then scored. On a 469-page audit that streamed 43,470 findings, the finalize step exceeded the hosting worker's 128 MB isolate and the run failed at the very last step, after the audit had already been paid for.

This replaces the materialization with an incremental fold. Findings arrive one page at a time and fold into per-rule `IssueTally` (the streaming-scoring machinery already used by the engine), and `calculateHealthScoreFromTallies` produces the score. Peak memory now tracks rules x checks instead of finding count.

## How

- `reconstructPageRuleChecks` reconstructs one page's checks, sharing the same inverse-of-`flattenChecks` code the whole-audit reconstruction uses.
- `foldCompleteStoreTallies` is the bounded twin of `reconstructCompleteResults` + `buildScoringResultsFromMerged`: the same synthetic-pass denominators, the same passing-sibling counts and their security clamp, the same carried replays.
- `runCloudSmartAudits` takes the findings as a page source and returns `scoringTallies`. In complete mode it flattens nothing, because the chunk ingest already persisted this run's rows with the store's earliest-first-seen conflict rule, and the caller passes the prior OPEN findings EXCLUDING this audit. That exclusion is a correctness requirement, and it is also what keeps the prior read bounded by the delta rather than by the whole store.
- `carriedFindingToCheck` is extracted so both scoring paths build a carried check from one construction.

## Why it is safe

Byte-identity is the gate. The parity harness now asserts the folded `HealthScore` equals the materialized one on every fixture, including:

- multi-item checks with a remainder, the shape that would diverge if a page were ever split across two folds (the per-key unit cap and the `additional` max are local to one fold call);
- the passing-sibling counts and their clamp;
- carried findings on un-crawled pages, covering the replayed carried checks and the clean-carried passes.

The page-boundary contract on the finding source is documented on the type and in the module header, because it is what the identity rests on.

## Measured

`scripts/bench-complete-store-fold.ts`, max RSS of a fresh process (Bun/JSC on a host, so the ratio transfers, not the absolutes):

| shape | materialized | bounded |
|---|---|---|
| 60,000 findings / 500 pages | 180.4 MiB | 53.0 MiB |
| 43,617 findings / 469 pages | 137.8 MiB | 51.6 MiB |

The bounded path stays flat as findings grow (20k 54.4 MiB, 60k 57.4 MiB, 150k 64.8 MiB) where the materialized one scales with them. Note the materialized peak at the real incident's size already exceeds 128 MB on the host, which is the failure being fixed.

## Tests

`bun test tests/complete-store-parity.test.ts` (23 pass), plus the merge, cloud-merge, advisory-scoring, streaming-scoring-golden and fingerprint-parity suites (40 pass). Typecheck clean.

### DIFF
--- a/packages/audit-engine/scripts/bench-complete-store-fold.ts
+++ b/packages/audit-engine/scripts/bench-complete-store-fold.ts
@@ -1,0 +1,147 @@
+// Complete-store finalize memory bench (#1873).
+//
+// Measures the two ways of scoring a chunked publish off the COMPLETE findings:
+//
+//   --mode=materialized  the pre-#1873 path: every page_findings row resident as a
+//                        PageFindingRecord[] (loadIngestedFindings), then
+//                        reconstructCompleteResults → buildScoringResultsFromMerged
+//                        → calculateHealthScore. This is what Cloudflare killed
+//                        with `exceededMemory` at 43,470 findings.
+//   --mode=bounded       the #1873 path: findings arrive one page at a time from a
+//                        cursor and are folded into per-rule tallies, then
+//                        calculateHealthScoreFromTallies.
+//
+// Peak memory is only honest as a fresh process's max RSS (a JSC heapUsed delta
+// reported a 4 MiB JSON.parse as +0.00 MiB — see the publish-zod-clamp
+// measurement), so run each mode under /usr/bin/time and compare `maximum
+// resident set size`:
+//
+//   /usr/bin/time -l bun run scripts/bench-complete-store-fold.ts --mode=materialized
+//   /usr/bin/time -l bun run scripts/bench-complete-store-fold.ts --mode=bounded
+//
+// Defaults reproduce the #1873 acceptance shape: 60,000 findings over 500 pages.
+// The absolute numbers are Bun/JSC on the host, not workerd/V8 in a 128 MB
+// isolate — the RATIO is what transfers.
+
+import type { CheckResult, PageFindingRecord } from "@squirrelscan/core-contracts";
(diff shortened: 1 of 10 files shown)


nc9:
Second commit fixes two defects a review pass found in the first, both of which move numbers a customer sees.

**Removed pages.** A page that returned 404/410 this run can still carry findings the crawl streamed before the status was known, and the materialized path dropped those checks from the union before scoring. The fold counted them: a site with one clean page and one deleted page scored 44 instead of 100. The fold now takes `removedUrls` and leaves a removed page's fresh checks out, exactly as the filter it replaces did. Pinned by a parity test that builds the same reference by hand.

**Occurrence totals.** `foldGroup` set `details.occurrences` to the number of objects it folded, which was correct only while every constituent was a single per-page check. The complete-store finalize re-folds the staged shell's aggregates together with carried checks, and one aggregate stands for hundreds of findings: 600 fresh represented by one aggregate plus 500 carried came out as 501 rather than 1100, and issue-sync persists that number as the tracker's occurrence count. Occurrences are now summed over the constituents' own totals. That is exact, unlike the `pagesTruncated` floor beside it, because `unfoldAggregateCheck` deletes `occurrences` when it expands an aggregate, so the unfold then re-fold path still counts one per page. Both directions are now tested in `fold.test.ts`.

PR #243, opened 2026-09-07
fix(utils): match identity by name AND value shape, and compose the two sorts (#242 round 3)

nc9:
Follow-up to #242, which was squash-merged from its 15:59Z head while the third codex round's fixes were still being pushed. This is that commit (0e7e720 from the original branch), cherry-picked onto main.

What it fixes, each with a fixture: deleting the Cloudflare block could manufacture a tag out of a stray `<` and a trailing attribute-looking string (a comment is left in place to keep the boundary); the two sorts overlapped so an app-block sort discarded the asset sort inside it (f(f(x)) now equals f(x)); only `rel="stylesheet"` links count as assets; adjacency uses HTML whitespace rather than `String.trim`; a close tag's end is found rather than assumed. Identity fields are matched by name AND value shape (reqid/requestId/eventMetadataId must hold a uuid, `u` exactly twelve lowercase hex), which needs no payload recognition: the payload-anchored variant on main passes unit tests but silently stops the corpus collapsing (15 bodies → 15 hashes) because one real rotating payload is a 16 KB minified bundle with no anchor.

Corpus (`packages/utils/scripts/fingerprint-corpus.ts`, 15 bodies over 35 min): 13 collapse to one hash, the 2 from before a theme publish keep their own.

Collision scope, accepted by the team lead: same domain, same URL, same tenant's own render, 7-day TTL. Deliberately malformed markup (`<plaintext>`, self-closed `<script/>`) that htmlparser2 tokenizes differently from a browser can only collide a page with another version of itself; that is out of scope for a Workers-bundled tokenizer.

Refs squirrelscan/repo#1899.

### DIFF
--- a/packages/utils/src/fingerprint.ts
+++ b/packages/utils/src/fingerprint.ts
@@ -73,19 +73,27 @@ const APP_BLOCK_BEGIN = /^\s*BEGIN app block:\s*shopify:\/\//i;
 const APP_BLOCK_END = /^\s*END app block\s*$/i;
 
 /**
- * Recognizes a Shopify analytics payload — the only place identity fields are
- * neutralized.
+ * Request/visitor identity fields, matched by NAME **and** by the shape of the
+ * value.
  *
- * Anchored on the payload, not on the field name alone and not on the uuid
- * shape. `"u"` in an arbitrary script can be content: the script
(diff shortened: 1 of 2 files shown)


nc9:
Superseded by #244 (same commit plus the file-header statement), opened by the lane that owns the change.

PR #268, opened 2026-09-08
docs(benchmarks): record the hosted crawl reuse control pair on our own site

nc9:
Records the hosted crawl reuse control pair under `## Hosted crawl reuse`, next to the drscholls rows.

The drscholls pair could not tell a broken feature from an origin unsuited to it. This is the same measurement against a site with the opposite property: our own marketing site, which serves `cache-control: public, max-age=60, s-maxage=600, stale-while-revalidate=86400` on every page. 150 pages, coverage full, render on, 22 minutes apart, on the code that charges adoption rather than lookup.

| | first run | second run |
|---|---|---|
| crawl | 742 s | 520 s |
| pages fetched | 150 | 114 |
| pages reused | 0 | 36 |
| credits | 350 | 302 |
| health score | 69 | 69 |

**Reuse fires and the double charge is gone.** 114 renders plus 36 adopted pages is exactly the 150 crawled, and the overlap between the two debit sets is zero, against 37 urls double-charged in the run above.

**The lower total is a shortfall, not a saving.** `render_cached` costs 2 credits like `render` by design, so a fully settled second run costs what the first cost. Two silent truncations are recorded and filed as squirrelscan/repo#1969 (settlement times out after ~12 pages) and squirrelscan/repo#1970 (the upload drops chunks with no event).

Docs only, no code changes. Reviewed with codex; both findings were over-claims in the prose and are fixed in the committed version.

### DIFF
--- a/benchmarks/2026-09-perf-program.md
+++ b/benchmarks/2026-09-perf-program.md
@@ -356,6 +356,74 @@ directives, and to pay nothing on this class until the per-request variation is
 characterised. That is what squirrelscan/repo#1899 acceptance criterion 1 was
 asking, and it is now answered: the body differs, the normalizer does not.
 
+### The control: an origin that does send freshness directives
+
+The drscholls pair could not tell a broken feature from an unsuitable origin, so
+the same measurement was repeated against a site chosen for the opposite
+property: our own marketing site, which serves `cache-control: public,
+max-age=60, s-maxage=600, stale-while-revalidate=86400` on every page. It has 64
+urls in its sitemap and the crawl reaches the 150-page cap through link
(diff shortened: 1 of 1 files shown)


nc9:
Provenance for the numbers, for anyone re-deriving them: the two runs are `01M1ZA52NM6YPG4NR47VGQNFEN` (seed, completed 01:45Z) and `01M1ZCAF7VHCFQSJ05ENN0RTNS` (measure, completed 02:19Z), both on the same internal org with identical config (150 pages, coverage full, render always, spend cap 350). Credits are summed from the ledger by run id, not read off an estimate; the render and adopted debit sets were compared for overlap directly.

PR #272, opened 2026-09-08
fix(cli): scope keys and auth whoami to a named org

nc9:
`squirrel keys create` resolved the org by taking the **first** entry of `GET /v1/organizations`, which is ordered newest-first. An account that gained a second org therefore started minting live credentials against whichever org was created last, while cloud audits kept spending the server's **active** org (`users.activeOrgId`). The two disagreed and neither was ever printed.

Verified read-only against production on a real two-org account: the membership list leads with the org created last, while the active org is the older one.

## What changed

- `keys create` / `list` / `revoke` take `--org <slug|id>`, an exact match on slug or id, never a prefix.
- `keys create` **refuses** to pick when the account has more than one org and no `--org`, listing them and marking the active one. Nothing is minted on that path.
- `keys list` groups by org and labels each. An org whose keys need a role you don't have is reported inline rather than failing the command; under `--json` that reason goes to stderr so the stdout array is never silently short.
- `keys revoke` searches every org and sends the DELETE to the org that owns the match. A prefix match while any org's listing failed is refused, since a prefix is only unique among the keys actually read; an exact key id is globally unique and still resolves.
- `auth status` / `auth whoami` print the active org's slug and id, resolved from the server rather than guessed, plus the org count. Best-effort: an unreachable API drops the org lines, not the command. The lookup keys on the token type, so an `sqcli_` login token supplied through the env var keeps its org lines.
- Shell completions for all three `keys` subcommands in bash, zsh and fish, with a drift test reconciling them against the citty definitions.

No API change: `GET /v1/organizations` already carries slug and role, and `GET /v1/hydrate` already carries `activeOrgId`.

## Docs

`docs/cli/keys.mdx` and `docs/cli/auth.mdx` document `--org`, the org lines, and the partial-listing behaviour. `make validate` passes.

## Verification

- CLI typecheck, oxlint and oxfmt clean.
- 146 tests across the touched files pass, including new suites for org resolution, cross-org listing, revoke search, and the active-org status lines.
- Smoke-tested against production read-only: `auth whoami` names the active org, `keys list` labels both orgs, `keys create` without `--org` refuses and mints nothing.

Refs squirrelscan/repo#1971


nc9:
**CI note: the two CodeQL alerts on this PR were line drift, not new findings.**

`js/clear-text-logging` fired at `apps/cli/src/cli/commands/keys.ts:117` and `:119`. The same rule is already dismissed as a false positive at five places in this file and four in `auth.ts` (alerts #31 to #39). My edit moved those statements about fourteen lines, the alert fingerprints stopped matching, and GitHub re-reported them as new.

What the flagged lines log is the key's **scope names** and its **prefix**, never the token. The taint source CodeQL names is `API_KEY_SCOPE_PRESETS`, a public enum of scope strings.

I dismissed both as false positives with that reasoning, matching the existing nine. Worth knowing: `code-scanning/alerts?state=open` returns zero for this repo, so it reads as "main is clean and this PR introduced them" until you page over every state.

All 13 checks are green. Not merging: this is for review.

PR #304, opened 2026-09-09
fix(crawler): give the entry URL a second chance and name the deadline when it times out (repo#1699)

nc9:
Fixes the second failure mode behind squirrelscan/repo#1699: a cloud audit of `nuxt.daigo.ru` (healthy origin, 200 on everything) failed with `No pages were crawled from nuxt.daigo.ru: The operation was aborted.` and reason code `unknown`. Root cause written up on the private issue; summary:

1. The site's sitemap lists 449 urls on the **sister domain** (`daigo.ru`), every one filtered as cross-domain, so the frontier is the seed alone and quick mode gates link discovery on that one fetch.
2. The seed fetch runs through the cloud document fetcher under the cloud's 12s per-request deadline and gets exactly one attempt (`fetchPageWithRetry`: `maxAttempts = options.fetcher ? 1 : 3`). Render and the plain fallback both hit the deadline.
3. `fetchWithDocumentFetcher` mapped the resulting Bun `AbortError` to `CrawlError.network(message)`, so the failure classified as `unknown` carrying the runtime's text. The standard path already mapped aborts to `timeout`.

## Changes (`packages/crawler`, `packages/core-contracts`)

- **Abort from a document fetcher is a timeout** that names the fetcher and the deadline (`page fetch via cloud-render gave up after 12000ms`); the runtime error code is forwarded for other transport failures, like the standard path. `CrawlError.timeout` takes an optional message; `crawlErrorToFailureDetail` keeps a non-generic timeout message as the detail.
- **The entry URL gets a second chance.** When its fetch times out (or fails transport-level through a document fetcher) and nothing has been stored yet, the crawl retries it once through a plain fetch with a doubled deadline and emits an `entry-fetch-retried` warning. A second timeout folds both deadlines into the reason (`…gave up after 12000ms; a plain retry gave up after 24000ms`), kept under the 120-char detail cap. A crawl that already has a corpus never pays this.
- **Preamble budget expiry says so.** Probes already degraded to the not-attempted shape (#194); `start()` now logs and emits a `preamble-budget-exhausted` warning naming the budget so a report whose AX probes are all unknown is explainable.
- `CrawlWarningCode` gains the two codes.

## Tests

`packages/crawler/tests/entry-fetch-timeout.test.ts` (6 tests): a synthetic Bun origin with a flat delay on every response, 200 + a large HTML body on every path (soft-404 shape), a 300-url cross-domain sitemap, and a document fetcher that aborts like the cloud one. Covers: abort → timeout classification and the exact reason sentence; slow-TTFB origin yields its entry page via the retry; aborting fetcher falls back to plain fetch and collects pages; an existing corpus never retries; nothing-fetchable names step and both deadlines and never says "The operation was aborted"; stalled probes crawl with one preamble warning.

Ran: `bun test` on the new file + root-failure-classification, preamble-budget, slow-origin-crawl, fetch-interrupt, crawler-stop-hard-cap, classifier-parity (all green); `tsgo --noEmit` in crawler, core-contracts, audit-engine, fetchers, apps/cli; `scripts/test-suites.ts --check`. Local run of the real crawler against nuxt.daigo.ru with the plain fetcher collects 10 pages in 12s.

Self-reviewed with a codex pass (`codex exec`, gpt-6-astra); findings addressed in follow-up commits on this branch where any.

No private-runtime change is required. Optional follow-up there: `CLOUD_CRAWLER.defaultTimeoutMs = 12_000` is the tightest bound in the system and also sizes the preamble budget and sitemap walk window.

Ref: squirrelscan/repo#1699

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,14 @@ How it works:
 
 ### Fixed
 
+- A cloud audit of a site whose sitemap points at another domain no longer fails
+  with "No pages were crawled from <site>: The operation was aborted." when the
+  entry page's first fetch hits its deadline. The entry URL now gets one more
+  plain fetch with a doubled deadline before the audit is given up (a `warning`
+  event says so), an abort from a document fetcher classifies as a timeout that
+  names the fetcher and the deadline instead of `unknown` with the runtime's
+  text, and a crawl whose preamble budget runs out records a warning naming the
+  budget. squirrelscan/repo#1699
 - `squirrel self update` now checks that the binary your PATH resolves is the one
   it just installed, and says so when it isn't. It used to flip the symlink
   recorded at install time and report success on that alone, so a stale
--- a/packages/core-contracts/src/index.ts
+++ b/packages/core-contracts/src/index.ts
@@ -785,7 +785,21 @@ export type CrawlerEvent =
  * probe cannot see an apex→www redirect the whole audit describes the wrong
  * host. The page fetch is the first evidence that contradicts it.
  */
-export type CrawlWarningCode = "seed-base-mismatch";
+export type CrawlWarningCode =
+  | "seed-base-mismatch"
+  /**
+   * The entry URL's fetch timed out before any page was stored, and the crawl
+   * is retrying it once through a plain fetch with a relaxed deadline
+   * (squirrelscan/repo#1699). The message names both deadlines.
+   */
+  | "entry-fetch-retried"
+  /**
+   * The crawl preamble's wall-clock budget ran out, so the root probes that
+   * had not started were skipped and the crawl went ahead with what it had
+   * (squirrelscan/repo#1699). Their records carry the not-attempted marker,
+   * never a confirmed absence.
+   */
+  | "preamble-budget-exhausted";
 
 export interface AuditLifecycleEvent {
   type:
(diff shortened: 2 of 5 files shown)


nc9:
Codex self-review (gpt-6-astra, medium) on the first commit returned 4 should-fix findings, all addressed in 14e7ad4:

1. The plain entry retry ran through `fetchPageWithRetry`'s three standard attempts at a doubled deadline and could outrun the per-URL watchdog. Now `FetchOptions.attempts: 1`, and the retry is skipped when the watchdog has less than the retry deadline (+1s) left.
2. `isAbortError` matched on the message ("aborted") as well as the name; a transport failure like "connection aborted by peer" would have become a claimed deadline. Name only now, and the detail reads "was aborted (deadline Nms)" because a fetcher may cut itself off before the deadline it was handed.
3. The double-timeout detail concatenated the first message ahead of the retry deadline, so the 120-char cap could drop the second number. It is now built from the two deadlines first: `entry page failed at 12000ms via cloud-render, then a 24000ms plain retry timed out`.
4. The test typed `rootFailure` narrower than `AuditFailureDetail` and read `.source` off it; fixed.

Re-ran: the new file plus root-failure-classification, fetch-interrupt, rate-limit-crawl, full-crawl-harness, preamble-budget, slow-origin-crawl, crawler-stop-hard-cap, seed-base-collapse, carried-seed-crawl (107 tests, green); tsgo in packages/crawler.

PR #305, opened 2026-09-09
fix(audit-engine): reserve plain-fetch headroom inside the cloud page deadline and raise it to 30s (repo#2026)

nc9:
Class fix under squirrelscan/repo#1699 (public #304 was the entry-URL second chance). Tracker: squirrelscan/repo#2026.

## The new deadline structure

Each request's `timeoutMs` is now the **outer deadline of the whole cloud fetch**, render and fallback included. Render has `outer - headroom` to itself. At that point, if nothing has landed, the plain-HTTP fallback starts with the remaining headroom and **races** the render; whichever settles first serves the page. The headroom is `CLOUD_CRAWLER.fallbackHeadroomMs` (12s, what the whole page fetch used to get), clamped to half the outer so render always keeps at least half. At the new cloud default of 30s: render alone for 18s, then 12s racing.

Consequences, each pinned by a test in `packages/audit-engine/tests/cloud-fetch-headroom.test.ts`:

- A render that never returns plus an origin that answers inside the headroom yields the page inside the outer deadline, even with a caller enforcing that deadline via `signal`. Before: the caller's abort landed first and the page was lost as "The operation was aborted."
- The fallback gets what is **left** of the outer deadline, never the request's full timeout again (render 10s + fallback 30s no longer happens).
- A render landing during the race, inside the deadline, still wins: paid work is used, one charge.
- A racing fallback that fails fast (refused/DNS) holds its failure until the outer deadline so a late render can still win.
- A per-url render error after the race started does not start a second fallback.
- **Charge path is abort-aware**: the submit is no longer cancelled by caller aborts (crawl stop / watchdog). The server debits on submit, so the old code could cancel the HTTP call mid-flight and leave a debit that nothing recorded (`onRenderCharged` never fired, `budget.spent` short by a batch). Now the submit runs to the batch budget and the debit lands in the accounting; aborted waiters still reject at once and never get a fallback.
- Through the real crawler: a wedged render on every page still collects every page, no `page:failed`, no `entry-fetch-retried`.

The in-flight render job is never cancelled by any of this, so its result lands in the server's render cache either way (a later render of the url is `render_cached`, same customer price).

Running the new test file against the old fetcher: 7 of 8 fail (the charge test fails on both charge assertions). After: 8 of 8 pass.

## The per-page deadline: 12s to 30s, with rationale

`CLOUD_CRAWLER.defaultTimeoutMs` 12s → 30s (max 30s → 60s), documented in `limits.ts` with the three derived windows:

| window | derivation | before | after |
|---|---|---|---|
| preamble budget | `min(45s, 3 × T)` | 36s | 45s |
| sitemap walk window | `min(20s, 3 × T)` | 20s | 20s |
| entry fetch | one outer deadline through the fetcher, then the #1699 plain retry at `2 × T` | 12s (+45s batch) +12s, retry 24s | 30s, retry 60s |
| per-URL watchdog | `max(120s, 6 × T)` | 120s | 180s |

Rationale against the run budget: T is a ceiling only a stalled page pays; the crawl-phase budget still bounds the run; the tail a stalled site can add is concurrency × 30s. On a quick run (130s crawl phase) with every stage at its bound: 45 + 20 + 30 = 95s before the entry retry, which the crawl-phase stop can cut short. The old path summed to 36 + 20 + 45 + 12 = 113s with no retry inside the phase, so this is not a regression there.

**Note for the cloud**: worker-agent's `component-runtime.ts` hardcodes `DEFAULT_CLOUD_CRAWLER_TIMEOUT_MS = 12_000` instead of importing `CLOUD_CRAWLER`; the constant change reaches the container only with that private 3-line swap (reported to the lead). The fetcher-level headroom structure applies at 12s too (render 6s alone, then 6s racing).

## Acceptance

- [x] Deadline structure reserves fallback headroom inside the outer per-fetch deadline; covered by a never-returning render + plain origin test where the page lands and no abort surfaces
- [x] Per-page deadline raised with a written rationale against the run budget; the three derived windows (preamble budget, sitemap walk, entry fetch) documented next to the constant
- [x] An aborted render leaves no charged-but-unrecorded render: the submit is never cancelled by a caller abort, the in-flight job is never cancelled, and a late render inside the deadline is used; charge on abort proven before (unrecorded) and after (recorded)
- [x] No behaviour change for the CLI plain path: `packages/crawler` fetch path untouched except a comment; golden and parity suites byte-identical (audit-engine `test:golden` set, crawler deadline suites, both cloud-fetcher suites all green)

## Verification

- `tsgo --noEmit` in core-contracts, crawler, audit-engine: clean
- audit-engine golden gate (22 files, chunks of 5): 161 pass, 0 fail
- crawler entry-fetch-timeout / slow-origin-crawl / preamble-budget / fetch-interrupt / deadline: 28 pass
- apps/cli `tests/crawler/cloud-fetcher.test.ts`: 38 pass; audit-engine `cloud-fetcher.test.ts`: 7 pass
- Codex review (gpt-6-astra, medium) run by the author; findings addressed or rebutted in a comment below.

No auto review bot on this repo; DCO signed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,16 @@ How it works:
 
 ### Fixed
 
+- Cloud rendering now reserves time for its plain-HTTP fallback inside each
+  page's fetch deadline: the render has the deadline minus 12 seconds to
+  itself, then the plain fetch starts and races it, and whichever lands first
+  serves the page. A render that finishes late but inside the deadline is still
+  used, and a crawl stop mid-submit no longer leaves the server's render debit
+  unrecorded. The cloud runner's per-page deadline rises from 12 to 30 seconds
+  (the CLI's default) so a far-away origin no longer decides the audit; the
+  preamble budget, sitemap walk window and entry-page retry that derive from
+  it are documented next to the constant. squirrelscan/repo#2026
+
 - A cloud audit of a site whose sitemap points at another domain no longer fails
   with "No pages were crawled from <site>: The operation was aborted." when the
   entry page's first fetch hits its deadline. The entry URL now gets one more
--- a/packages/audit-engine/src/cloud-fetcher.ts
+++ b/packages/audit-engine/src/cloud-fetcher.ts
@@ -18,6 +18,21 @@
 // 3 consecutive 5xx/transport failures) the fetcher flips PERMANENTLY to the
 // provided plain-HTTP fallback fetcher (announcing via onFallback once); a
 // one-off render failure/timeout falls back for that url only.
+//
+// DEADLINE STRUCTURE (squirrelscan/repo#2026): each request's `timeoutMs` is the
+// OUTER deadline of the whole fetch, render and fallback included. Render has
+// `outer - headroom` to itself; at that point, if nothing has landed, the plain
+// fallback starts with the remaining `headroom` and RACES the render, and
+// whichever settles first serves the page. So the fallback is guaranteed its
+// headroom inside any outer deadline (before this, the render machinery ran
+// under a 45s batch budget of its own and a caller enforcing a shorter deadline
+// aborted the waiter before the fallback ever dispatched), and a render that
+// lands late but inside the deadline is still used rather than discarded — it
+// was charged on submit. The in-flight render job is never cancelled: its
+// result lands in the server's render cache either way. A caller abort (the
+// crawler's watchdog / stop) still rejects without a fallback, but no longer
+// cancels the SUBMIT: the server debits on submit, so the submit runs to its
+// own bound and the debit is recorded whether or not anyone still wants the page.
 
 import type {
   RenderChargeLine,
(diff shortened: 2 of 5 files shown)


nc9:
Codex review (gpt-6-astra, medium) on the fetcher diff: 3 defects reported, triage:

1. **Headroom timer armed at batch start, not at accept** (real). Fixed in the follow-up commit: `armRace` runs in `enqueue`, so the coalescing window and a slow submit cannot eat the fallback's slice. Test: a coalescing window longer than the render's share, page still lands inside the outer deadline.
2. **Last racing waiter settling during a poll sleep did not stop polling** (real, one extra round trip). Fixed: every settle runs a batch hook that aborts the poll signal once nothing live remains; polls are also bounded by the batch deadline. Test: poll count after settlement is at most one in-flight poll.
3. **A failed submit response does not prove no debit occurred** (rebutted as out of scope). This is pre-existing: the old code released the reservation on any submit error too. Closing it needs a server-side idempotent submit id to reconcile against; holding "uncertain" reservations client-side would leak budget forever on every transport error. This PR narrows the window (caller aborts no longer cancel the submit); the residual is the transport-loss case, tracked in the private tracker note.

Risks it listed, accepted: the render winner does not cancel the competing plain fetch (bounded by the headroom); a racing blocked render loses the `render-block` fallback reason tag (observability only); non-finite option values are not validated (callers pass constants).

PR #311, opened 2026-09-09
docs(mcp-clients): outcome-first title naming the top clients

nc9:
`/developers/mcp-clients` is the most-shown docs URL in Google Search Console and converts nothing: **121 impressions, 0 clicks, 0.0% CTR, average position 12.0** over 2026-09-03 to 09-10. The queries in that window are of the "codex mcp" / "cursor mcp" / "cursor tool" kind, and the old title never said what the server does or named a single client.

Same shape as the `/for/codex` fix in the private repo (#1842 / #1843).

## Change

| | before | after |
|---|---|---|
| `title` | `MCP client setup` (16) | `Connect Claude Code, Cursor, or Codex to website audit MCP` (58) |
| `description` | 131 chars, led with "Exact config for connecting" | 147 chars, leads with "Copy-paste config that connects" |
| `sidebarTitle` | absent | `MCP client setup` |

The title leads with the connect action, names the top clients including the two the queries actually ask for, and says what the server does. It is 58 characters, under the 60-character SERP truncation point.

The new `sidebarTitle` keeps the nav label at "MCP client setup", so the sidebar and the dozen or so cross-page links and cards that read "MCP client setup" all still agree with what the nav renders. `sidebarTitle` is already the pattern on every page under `developers/agents/`.

## Validation

`make validate` (`tangly check --strict`) passes: 392 pages, all checks passed. The 1 reported orphan is pre-existing, confirmed by running the same check against the unmodified file on `main`.

Body copy is untouched. Frontmatter only.

Refs https://github.com/squirrelscan/repo/issues/2049


nc9:
**CI note: the one red check is not from this PR.**

`Dependency audit` fails on a newly published advisory, not on anything in this diff. This branch changes three lines of frontmatter in one `.mdx` file and touches no `package.json` and no `bun.lock`.

```
smol-toml@1.7.0
  workspace:@squirrelscan/cli > smol-toml
  workspace:@squirrelscan/docs > tangly > ... > @astrojs/internal-helpers > smol-toml
  high: smol-toml: Denial of Service via malformed TOML documents (<=1.7.0) - GHSA-7w5x-hrqm-74c2
```

It will fail the same way on `main` and on every other open PR until the dependency is bumped. Patched `1.8.0` is published and the declared ranges (`^1.6.1`, `^1.6.0`) already allow it. Tracked separately in https://github.com/squirrelscan/repo/issues/2053 rather than fixed here, to keep this a frontmatter-only change.

Everything else is green, including `Docs check and build` and `Verify commit sign-offs`.

PR #313, opened 2026-09-09
build(deps): bump smol-toml to 1.8.0 for GHSA-7w5x-hrqm-74c2 (repo#2053)

nc9:
Clears advisory [GHSA-7w5x-hrqm-74c2](https://github.com/advisories/GHSA-7w5x-hrqm-74c2) (`smol-toml <=1.7.0`, denial of service via malformed TOML), which has been failing the `Dependency audit` job on every PR since 2026-09-09.

Refs squirrelscan/repo#2053

## Why the obvious fix does not work

`bun update smol-toml` resolves the direct `apps/cli` dependency but leaves the transitive astro chain nested at 1.7.0, and it also adds a spurious `smol-toml` entry to the root `dependencies`. The audit stays red, and reports it as a direct dependency on top:

```
smol-toml  <=1.7.0
  (direct dependency)
  workspace:@squirrelscan/cli > smol-toml
  workspace:@squirrelscan/docs > tangly
```

`astro` and `@astrojs/internal-helpers` both declare `^1.6.0`, so nothing forces them onto the patched line.

## What this does instead

- Pins `smol-toml: ^1.8.0` in the root `overrides` block, which is what this repo already uses for exactly this pattern (16 other entries).
- Lifts the CLI's declared floor from `^1.6.1` to `^1.8.0`, so the reachable direct dependency cannot drift back if the override is ever removed.
- Regenerates `THIRD_PARTY_NOTICES.md` and `npm/THIRD_PARTY_NOTICES.md`, which CI's `notices:check` gates on. Only the version string moves; the license stays BSD-3-Clause.

`bun.lock` now carries exactly one resolution, and no nested `<pkg>/smol-toml` entries remain.

No new `--ignore` was added. `audit:dependencies` still carries the same eight.

## Verification

| Check | Before | After |
|---|---|---|
| `bun run audit:dependencies` | 1 high, exit 1 | clean, exit 0 |
| `bun run notices:check` | exit 0 | exit 0 (after regen) |
| `bun run docs:check` | | exit 0, 392 pages |
| `bun run docs:build` | | exit 0, 395 pages |
| `apps/cli` `tsgo --noEmit` | | exit 0 |
| CLI suite (`test-suites.ts --run cli`) | | 2133 pass, 1 skip, 0 fail, 165 files |

`smol-toml` is only used for `parse`/`stringify` of `squirrel.toml` in four CLI files. Behaviour relied on by `apps/cli/src/config.ts` is unchanged on 1.8.0: a `[__proto__]` table still becomes a real own key rather than polluting `Object.prototype`, so the `deepMerge` guard's documented premise still holds. `tests/security/prototype-pollution.test.ts` covers it and passes.


nc9:
Review pass (codex, plus independent re-verification) found no regression from this diff. Two behaviour notes, neither blocking:

**1. `stringify` of integers beyond `Number.MAX_SAFE_INTEGER` gained a `.0` suffix.**

```
parse+stringify "a = 9007199254740992"
  1.6.1 -> a = 9007199254740992
  1.8.0 -> a = 9007199254740992.0
```

Not reachable here. Every numeric field in `packages/config/src/schema.ts` is a bounded count, timeout, or concurrency limit, and safe-range integers stringify unchanged (`a = 1000` stays `a = 1000`). The CLI only calls `stringify` to write `squirrel.toml` in `init` and `config`.

**2. Pre-existing dead branch, not caused by this bump.** `apps/cli/src/config.ts:177` tests `error.name === "TomlError"`, but `smol-toml`'s `TomlError` never sets `name`, so it is the inherited `"Error"` on both 1.6.1 and 1.8.0. The friendly "TOML syntax error:" message has never fired. Filed as squirrelscan/repo#2055 and deliberately left out of this PR to keep the security bump minimal.

Also verified unchanged on 1.8.0: date and offset round-trips, `TomlError` thrown for malformed input, and the `[__proto__]`-becomes-an-own-key premise that `deepMerge` in `apps/cli/src/config.ts` documents and guards against.

PR #325, opened 2026-09-12
fix(installer): fall back to install.squirrelscan.com when GitHub cannot be reached (repo#2064)

nc9:
Fixes the private tracker's squirrelscan/repo#2064. Pairs with the worker route in squirrelscan/repo#2065, which serves the mirror.

## The failure

A GitHub release-asset URL redirects to githubusercontent.com (`objects.*` historically, `release-assets.*` today), and some networks cannot reach it. The manifest fetch from `install.squirrelscan.com` succeeds because that is our own origin, so the install gets as far as knowing which binary it wants and then fails to download it, on every attempt, forever. Sentry INSTALLER-WORKER-5: ten events since July, linux/x86_64, `Asia/Shanghai`, one user retrying twice within nine minutes.

## What changed

Every release asset now has two sources. GitHub first, then the same file from `install.squirrelscan.com/dl/<version>/<asset>`, which proxies it through Cloudflare's edge and reaches GitHub even when the client cannot. The manifest's `sha256` is verified over whatever arrives either way, so the second source adds no trust: a mirror is not a shortcut around the integrity check.

- `SQUIRREL_FORCE_MIRROR` skips github.com from the start. Value semantics, matching `SQUIRREL_NO_UPDATE`: empty, `0` and `false` are off. Presence semantics would let a stray `SQUIRREL_FORCE_MIRROR=` in a profile silently reroute every install. `SQUIRREL_DOWNLOAD_ENDPOINT` moves the mirror, for testing.
- **Both the manifest and the binary** go through the fallback. A fully blocked host fails at the manifest first, so covering only the binary would have left the reporting user exactly where they started.
- When both sources fail, one line names the hosts and the escape hatch, then a block prints the two URLs, the retry recipe and a by-hand route. The report keeps its existing step, `download_manifest` or `download_binary`, and carries both URLs in `error_output`: `error_line` is capped at 200 chars and two asset URLs are about 170.
- `install.ps1` gets the same two sources.

## Safe to ship before the worker route exists

`/dl` is not deployed yet. A worker without the route answers 404, `curl -f` treats that as a failure, and the script reports the GitHub failure it would have reported anyway. Nothing regresses for the overwhelming majority whose GitHub download works: they never reach the second source.

## Evidence

`curl` is shimmed to blackhole github.com the way the reporting network does, and to serve the real v0.0.94 darwin-arm64 binary and manifest from the mirror path.

A source that answers 200 with a sign-in page, which is what a captive portal or an intercepting proxy does, counts as a failed source: the installer moves to the other host rather than stopping on a parse error.

**GitHub blocked, mirror reachable** (acceptance box 1 and 2):

```
==> Downloading manifest...
Warning: github.com could not serve the manifest, trying install.squirrelscan.com...
==> Downloading squirrel v0.0.94...
Warning: github.com could not serve the binary, trying install.squirrelscan.com...
:: Downloaded from install.squirrelscan.com
==> Verifying checksum...
:: Checksum verified: 9ccfb658e573eda7...
==> Running self install...
✓ squirrel v0.0.94 installed
--- install.sh exit: 0
--- <bin_dir>/squirrel --version: 0.0.94
```

The shim saw both hosts, in order:

```
https://github.com/squirrelscan/squirrelscan/releases/download/v0.0.94/manifest.json
https://github.com/squirrelscan/squirrelscan/releases/download/v0.0.94/squirrel-0.0.94-darwin-arm64
https://install.squirrelscan.com/dl/v0.0.94/manifest.json
https://install.squirrelscan.com/dl/v0.0.94/squirrel-0.0.94-darwin-arm64
```

**Both hosts blocked** (acceptance box 3):

```
Error: Failed to download the binary from github.com and install.squirrelscan.com (retry with SQUIRREL_FORCE_MIRROR=1 to skip github.com)
  Tried:
    github.com               https://github.com/squirrelscan/squirrelscan/releases/download/v0.0.94/squirrel-0.0.94-darwin-arm64
    install.squirrelscan.com https://install.squirrelscan.com/dl/v0.0.94/squirrel-0.0.94-darwin-arm64

  If github.com is blocked on this network, skip it and retry:
    curl -fsSL https://install.squirrelscan.com | SQUIRREL_FORCE_MIRROR=1 bash

  If both hosts are blocked, download squirrel-0.0.94-darwin-arm64 on a machine that can
  reach one of them, copy it here, then:
    chmod +x squirrel-0.0.94-darwin-arm64 && mkdir -p '<bin_dir>' && mv squirrel-0.0.94-darwin-arm64 '<bin_dir>/squirrel'
--- install.sh exit: 1
```

What reaches Sentry:

```json
{"script":"sh","script_version":"2","os":"darwin","arch":"arm64","step":"download_binary","exit_code":1,
 "error_line":"Failed to download the binary from github.com and install.squirrelscan.com (retry with SQUIRREL_FORCE_MIRROR=1 to skip github.com)",
 "error_output":"tried https://github.com/.../squirrel-0.0.94-darwin-arm64 (failed) tried https://install.squirrelscan.com/dl/v0.0.94/squirrel-0.0.94-darwin-arm64 (failed)"}
```

**`SQUIRREL_FORCE_MIRROR=1`**: github.com is never contacted, the install completes, `squirrel --version` reports 0.0.94.

## CI now executes install.ps1, not just parses it

The install-test workflow fetches the published script from `main`, so nothing ever ran a change to `install.ps1` before users did; CI parsed it and stopped there. `scripts/install-ps-contract.test.ps1` dot-sources the script above `Main`, stubs `Invoke-WebRequest`, and drives the two-source path on the ubuntu runner's pwsh. 27 checks.

It earned its place immediately: it caught a regression in this very change, where hoisting the bin directory to script scope made `Join-Path` throw on a missing `LOCALAPPDATA`, which under `$ErrorActionPreference = "Stop"` would have killed the installer on line one, before the banner.

It also surfaced that a failed `Invoke-WebRequest -OutFile` leaves a partial file behind, which the mirror leg would then have been asked to overwrite. The failed attempt is deleted now.

## Review

Three rounds. The findings and how each was handled are in a comment below; the short version:

An acceptance pass returned FIX FIRST with five items, all applied: the redaction had to drop query and fragment as well as userinfo; a 200 carrying HTML had to count as a failed source rather than dying at jq; the PowerShell recipe had to escape apostrophes; the source line had to appear only for the mirror so the GitHub path stays byte-identical; and both CI steps had to probe for pwsh rather than declare it as the shell.

My own round on that delta found three more, including the blocker reopening itself: the redaction ran the userinfo rule before stripping the query, and that pattern crosses a `?`, so a token in the query survived into both the guidance and the JSON payload. It also found that the jq and no-jq manifest validators disagreed with each other, and that PowerShell accepted a scalar `binaries`.

An earlier codex pass (`gpt-6-astra`, read-only) returned 3/5 with six findings. Five were real and are fixed in 50f9b63; the sixth is the deliberate cross-repo split.

1. **P1, the mirror is not deployed yet, so this alone does not fix the issue.** Correct and by design. The worker route is squirrelscan/repo#2065, and shipping order does not matter: see the section above. Called out so the release that carries this is not mistaken for the fix landing.
2. **P2, `SQUIRREL_DOWNLOAD_ENDPOINT` credentials could reach the reporting endpoint.** Real. A mirror set to `https://user:token@host` put that token in `error_output` verbatim; the scrubber handles home paths and length, not URL userinfo. Both installers strip `user:password@` before printing or reporting, and leave an ordinary URL and an `@` inside a path alone.
3. **P2, the Windows recipe moved a file into a directory that does not exist.** Real. `Move-Item` does not create parents and the bin directory is absent before the first successful install. It creates the destination first now.
4. **P2, the bash recipe broke on a path with a space.** Real. The bin directory was interpolated unquoted. Single-quoted now with embedded quotes escaped, and a test round-trips a path holding a space, a quote and a dollar back through the shell as exactly one argument.
5. **P2, the one-line failure did not name the escape hatch.** Fair reading of the criterion. The line now carries `SQUIRREL_FORCE_MIRROR=1`, and says "github.com was skipped" instead when the variable is already set. Both forms stay inside the 200-char cap.
6. **P3, a skipped GitHub leg was reported as failed.** Real. With the flag set we never asked GitHub, and a report saying otherwise sends whoever reads it after the wrong host. Skipped and failed are distinct now.

Codex also confirmed bash 3.2 syntax, that the single EXIT trap survives, that both URLs and the step reach the EXIT handler without subshell variable loss, that every executed curl keeps its transport hardening, and that checksum verification is intact on both paths.

## Acceptance criteria (squirrelscan/repo#2064)

- [x] `curl -fsSL https://install.squirrelscan.com | sh` completes where github.com and the githubusercontent asset host are unreachable, by falling back to a download host under squirrelscan.com. Transcript above; needs squirrelscan/repo#2065 deployed to work against the real mirror.
- [x] The fallback serves the same bytes and the installer still verifies the checksum from the release manifest. The route is a byte-for-byte proxy and the checksum step is unchanged; the transcript shows it verifying after a mirror download.
- [x] When both downloads fail, one actionable line names the hosts tried and the variable that forces the mirror, and `installer_step = download_binary` is reported with both URLs. Transcript and JSON above.
- [x] `install.ps1` gets the same fallback. Plus 27 behavioural checks now running in CI.

## Verified end to end against production

The worker route is deployed, so this is a real install rather than a simulated one. `curl` is shimmed to refuse `github.com`, `*.githubusercontent.com` and `api.github.com`, and to hand everything else to the real curl:

```
==> Downloading manifest...
Warning: github.com could not serve the manifest, trying install.squirrelscan.com...
==> Downloading squirrel v0.0.94...
Warning: github.com could not serve the binary, trying install.squirrelscan.com...
:: Downloaded from install.squirrelscan.com
==> Verifying checksum...
:: Checksum verified: 9ccfb658e573eda7...
Installed v0.0.94
--- squirrel --version: 0.0.94
```

A 72MB binary pulled through the mirror, checksum verified against the real release manifest, installed binary runs. The route itself answers 200 with the right content-type and length for the manifest and the binary, 206 with a correct `content-range` for `Range: bytes=0-99`, and 404 for a version that does not exist.

Before the route was deployed it answered 404, which `fetch_with_retry` already treats as a failed attempt, so the ordering was safe either way.

## Checks run

```
bun test scripts/install-contract.test.ts          108 pass, 0 fail
pwsh -File scripts/install-ps-contract.test.ps1     68 checks, all pass
bash -n install.sh                                  clean
shellcheck -S warning install.sh                    clean
actionlint .github/workflows/ci.yml                 clean
detect-secrets against every tracked file           clean
bun run lint / typecheck / format:check             clean
cd docs && make validate                            393 pages, all checks passed
```

The PowerShell runs were done locally in `mcr.microsoft.com/powershell` since this machine has no pwsh; CI runs the same file on the ubuntu runner. Redaction, quoting, both manifest validator paths and a full install over the mirror were also verified under `/bin/bash` 3.2, the bash macOS actually ships.


nc9:
Verified the "safe to ship before the worker route exists" claim against the live worker rather than leaving it as an assertion:

```
$ curl -s -o /dev/null -w '%{http_code}\n' https://install.squirrelscan.com/dl/v0.0.94/manifest.json
404
$ curl -s -o /dev/null -w '%{http_code}\n' https://install.squirrelscan.com/dl/v0.0.94/squirrel-0.0.94-linux-x64
404
$ curl -fsSL -o /tmp/probe https://install.squirrelscan.com/dl/v0.0.94/manifest.json
curl: (56) The requested URL returned error: 404
exit: 56
```

A non-zero curl is what `fetch_with_retry` already treats as a failed attempt, so against today's production the mirror leg is a no-op and the script reports the GitHub failure it would have reported anyway. Nothing regresses for the majority whose GitHub download works: they never reach the second source at all.

Both codes go green once squirrelscan/repo#2065 is deployed, which is what actually closes the issue.

Also fixed two CI gates in 673efec, both firing on the synthetic credential the new redaction tests exist to exercise: detect-secrets reads `https://user:pass@host` as Basic Auth credentials (including inside a comment, so the two shipped scripts now describe the case in words and carry no credential-shaped literal at all), and CodeQL flagged `.includes("github.com")` as incomplete URL substring sanitization, which is a fair shape to flag even in a test. That assertion names the exact URL now.

nc9:
## Acceptance findings, all five applied (8f44f6a), plus my own round on the delta (05b6126)

### The five from the acceptance pass

1. **Redaction reduces a URL to scheme, host and path** in both scripts, so a token in the userinfo, the query or the fragment is gone before it reaches the telemetry POST or the printed recipe. A contract case asserts the JSON payload that actually leaves the machine for `https://u:pw@host/x?token=y#z`.
2. **A 200 carrying HTML is a failed source.** Each attempt validates the manifest before accepting it, so a bad first source falls through and exhaustion routes through the actionable error. Previously this exited 5 on a jq parse error with an empty report. The jq extraction needed its own guard too: a command substitution carries the failure to the assignment under `set -e`. PowerShell validates the parsed object's shape, since `Invoke-RestMethod` hands back an HTML body as a string rather than throwing.
3. **PowerShell recipe quoting.** Both arguments go through a helper that doubles embedded apostrophes. A test parses the printed command and round-trips its arguments for a path under `O'Brien`. `Join-Path` was wrong here: it resolves drives and rejects a `C:\` path on any non-Windows host.
4. **The source line appears only when the mirror answered.** Verified with a shim mode where GitHub works: the GitHub path prints no new line at all.
5. **One guarded CI step**, probing with `command -v pwsh` from bash and skipping with a notice. The full-file parse moved into the contract script, which also removed the inline PowerShell that shellcheck was flagging through actionlint.

### My own codex round on that delta found three more, one of them the blocker reopening itself

- **P1, the redaction order leaked.** The userinfo rule ran first and `[^/@]*@` crosses a `?`, so `https://mirror.test?token=user@querysecret/path` became `https://querysecret/path`: it ate the query delimiter and promoted the token's value to the host. Reproduced in both the printed guidance and the JSON payload. Query and fragment are cut first now, and the userinfo pattern is bounded so it cannot cross a delimiter even if that order is changed back.
- **P2, the two manifest validators disagreed**, which means a machine without jq got a different verdict on the same bytes. HTML carrying `{"version":...}` in a script tag passed the grep path, and a valid manifest that wrapped after `"binaries":` failed it, because grep is line-based. Rejecting a good manifest is the worse half. The no-jq path now requires the first non-whitespace byte to be `{` and matches the flattened text; the jq path requires `version` too. Both agree on all three fixtures now.
- **P2, PowerShell accepted `{"binaries":"unavailable"}`**, skipping a healthy mirror and reporting "No binary for platform" from a source that plainly failed. `binaries` must be a map.

Codex also correctly noted that my "GitHub HTML falls through" test was simulating a transport failure rather than the scenario it named. The shim takes a second body for URLs matching a needle now, so GitHub actually answers 200 with HTML while the mirror serves a real manifest, and the test asserts GitHub was called exactly once.

### Checks

```
bun test scripts/install-contract.test.ts          108 pass, 0 fail
pwsh -File scripts/install-ps-contract.test.ps1     68 checks, all pass
bash -n / shellcheck -S warning                     clean
actionlint                                          clean
detect-secrets                                      clean
bun run lint / typecheck / format:check             clean
cd docs && make validate                            393 pages, all checks passed
```

Also verified under `/bin/bash` 3.2, the bash macOS actually ships: redaction, quoting, both validator paths, and a full install over the mirror.

nc9:
## Real end-to-end install against production, GitHub blackholed

The mirror is deployed now, so acceptance box 1 is no longer simulated. `curl` is shimmed to refuse `github.com`, `*.githubusercontent.com` and `api.github.com` and to hand everything else to the real curl, so these are live bytes from install.squirrelscan.com.

```
==> Installing squirrel...
==> Detected platform: darwin-arm64
:: Fetching releases (channel: stable)...
==> Latest version: v0.0.94 (channel: stable)
==> Downloading manifest...
Warning: github.com could not serve the manifest, trying install.squirrelscan.com...
==> Downloading squirrel v0.0.94...
Warning: github.com could not serve the binary, trying install.squirrelscan.com...
:: Downloaded from install.squirrelscan.com
==> Verifying checksum...
:: Checksum verified: 9ccfb658e573eda7...
==> Running self install...
Installed v0.0.94

--- installed binary ---
squirrel -> …/.squirrel/releases/0.0.94/squirrel
0.0.94
```

A 72MB binary pulled through the mirror, checksum verified against the real release manifest, and the installed binary runs. The whole reason the reporting user could not install.

The route itself, checked directly:

| request | result |
| --- | --- |
| `GET /dl/v0.0.94/manifest.json` | 200, `application/json; charset=utf-8`, 1509 bytes |
| `HEAD /dl/v0.0.94/squirrel-0.0.94-linux-x64` | 200, `application/octet-stream`, content-length 90903752, `accept-ranges: bytes` |
| `GET` with `Range: bytes=0-99` | 206, `content-range: bytes 0-99/90903752`, 100 bytes returned |
| `GET /dl/v9.9.9/squirrel-9.9.9-linux-x64` | 404 |

The content-length matches the size the manifest records for that asset, and the 404 is what makes this safe against an older worker.

Also fixed the CodeQL alert on the last push: `a.includes("github.com")` in a test assertion is flagged as incomplete URL substring sanitization, fairly, so both call-count assertions now match the exact URL the script builds.

PR #336, opened 2026-09-13
feat(cli): squirrel entities command with filters, export formats and --diff (#2092)

nc9:
Implements `squirrel entities` (squirrelscan/repo#2092, epic #2061 sections 5 and 9) on top of the entity map that #333 made every audit build.

## The command

Shaped like `squirrel report`: same flag names, same project-scoped store access, same `--input` escape hatch.

| Invocation | What it does |
|---|---|
| `squirrel entities` | Summary of the latest audit's map |
| `squirrel entities <@id\|key\|name>` | One entity in full: properties, declaring pages, references, conflicts |
| `squirrel entities --list` | Stored audits with their entity counts and stable-`@id` share |
| `squirrel entities --diff` | What changed between two audits |

`--type`, `--page` and `--problem` produce a filtered **map**, not a filtered view, so `-f graphml --type Organization` hands a graph tool the filtered graph and the summary counts describe what survived rather than the whole site. `--problem` accepts `no-id`, `conflict`, `dangling`, `single-page` and `split-identity`.

Exit code is 0 on every path that produced an answer, including a site that declares no JSON-LD at all. That is a finding, not a failure. Only a usage mistake or an unreadable store exits 1.

## Export formats

`csv`, `dot`, `graphml` and `mermaid` join the existing `json`, `jsonld`, `html` and `markdown`, all in `packages/report` beside the renderers they sit with.

Every string in a map comes from an audited page, so each renderer escapes for its own parser rather than sharing one escape: RFC 4180 quoting plus a leading-quote guard against spreadsheet formula injection for csv, backslash escaping for dot, XML 1.0 character-range filtering (control bytes, U+FFFE/U+FFFF and unpaired surrogates are dropped, not escaped, because `&#x1;` is itself invalid) for graphml, and label sanitisation with quoted edge labels for mermaid. Mermaid stops rendering usefully past a few hundred nodes, so it draws the 150 most-declared and says so in a comment.

## `--diff`

New `diffEntityMaps` in the engine, exported from the `@squirrelscan/audit-engine/entity-map` subpath.

The property it exists to protect: two audits of a site rarely cover the same pages, and a naive set difference reports every entity on a page that simply was not visited as removed, which reads as "the site deleted its structured data". So an entity is only **removed** when every page that declared it was crawled again, and everything else lands in **not crawled again**.

Proving that is the subtle part. A node's `pages` is capped at 50 in the document with `morePages` counting the rest, so deciding on the sample calls a 101-page entity removed the moment the first 50 are recrawled. Coverage is therefore taken from the uncapped per-page `declares` index, and only when that index is itself complete (a publish projection clips it, which the crawl's own page count detects). Where coverage cannot be proven, the answer is not-crawled: conservative, says less, never the reverse.

The same rule governs fixes. A conflict or dangling reference counts as resolved only when the entity survived **and** all of its declaring pages were visited again, because a conflict needs two declarations to exist and seeing one of them tells you nothing.

Identity matching across an `@id` change pairs one-to-one: a group of same-named candidates is only matched when exactly one unmatched candidate exists on each side, so an ambiguous group falls through to added/removed rather than inventing a rename and a deletion at once. Signatures are `JSON.stringify([sortedTypes, name])`, never a delimiter join, since `@type` is site-controlled.

## Review

Two `codex exec` passes over the non-test diff. The first found 16 correctness issues, the second confirmed 12 of those fixes and found the remaining 4 gaps plus one new defect. All are now fixed, with tests for the ones that were behavioural.

The load-bearing ones:

- Both coverage defects above, and one-to-one identity matching.
- `getCrawlPageUrls` used request URLs and every status where the collector uses final URLs and 2xx only, so a redirect became a page no entity row could name.
- `backfillDerivedFields` repaired present-but-invalid values into a document that then validated. Absent means an older build did not write the field; present-and-invalid means the file is broken.
- A store that could not be **read** exited 0 with its error on stdout, exactly like "no audit stored yet". Those call for opposite next steps.
- `--diff` looked for the same-site predecessor inside the 50 most recent audits, but that window is global and ordered by time, so a busy site buried a quiet site's previous audit.
- `-o` wrote whatever `fmt` produced, and `fmt` decides colour from whether **stdout** is a TTY. Run from a terminal, the entity and list views wrote escape bytes into the file.

## Verification

- `tsgo --noEmit` clean in core-contracts, report, crawler, audit-engine and cli
- 117 tests pass across the entity-map suites: 25 diff, 16 export, 18 filter, 6 loader, plus the existing builder, streaming, slim, html and markdown suites
- `oxfmt` and `oxlint` clean on `apps/cli`
- `make validate` clean in `docs/` (394 pages, no broken links)
- End to end twice over. Against the real before/after snapshots of docs.squirrelscan.com, committed as fixtures: every format, every filter, and each error path checked for its exit code. Then against a live project store, which is where the `--list` column overflow and the empty-older-side case surfaced.

## Docs

New `/cli/entities` reference page, added to the CLI nav and the command table.

Closes squirrelscan/repo#2092

https://claude.ai/code/session_01T8w822GqnrviVMytnAsKFP


nc9:
Follow-up filed as squirrelscan/repo#2105: three fields a map reassembled from the project store cannot recover exactly (a page's declaration count and references, an edge's pages), plus the fact that the store does not record whether an audit wrote a map at all. None is load-bearing here, and this PR documents each as an approximation rather than emitting a plausible number. Fixing them needs an additive migration to tables #2091 had only just shipped, which seemed worth separating from this change.

https://claude.ai/code/session_01T8w822GqnrviVMytnAsKFP

Issue #344, opened 2026-09-13
docs: /agent-setup/prompt.md is a dead link on the agent setup page


nc9:
The agent setup page tells a user to paste a prompt that fetches `https://docs.squirrelscan.com/agent-setup/prompt.md`, and the CopyPrompt button links to `/agent-setup/prompt.md`. No such file exists in `docs/`.

`docs/agent-setup/index.mdx` lines 21 and 25.

`tangly check --strict` passes, because it treats an id it does not recognise as an external route rather than a dead link. A filesystem check over the .mdx tree plus the static files under `docs/` finds it: it is the only dead internal link in 2,785 across the whole docs tree.

This is the one page where a dead link costs the most, since the instruction is for an agent to fetch that exact URL, and the failure is silent from the user's side.

Either the file was never committed or it is expected to be generated at build time and is not. Found while adding the entity map docs pages; not fixing it there because I do not know which of the two it is.

nc9:
The link works. `docs/agent-setup/prompt.mdx` (added in #131) is served as markdown at https://docs.squirrelscan.com/agent-setup/prompt.md (HTTP 200, `text/markdown`, the page's own content), and a non-existent `.md` path returns 404. The filesystem check looked for a literal `prompt.md` file, so it missed the `.mdx` source.

PR #346, opened 2026-09-13
docs(entity-map): the REST endpoint answers 501 for four formats

nc9:
Follow-up to #345, from the cloud lane reading the merged page against their implementation.

`cloud.mdx`'s API table listed all eight export formats for `GET /v1/reports/{id}/entities` with no note that half of them do not render there. `csv`, `dot`, `graphml` and `mermaid` answer **`501`**, "The requested export format is not implemented in this release". Only `json`, `jsonld`, `html` and `markdown` come back. A reader following that table hit an error the page did not mention.

The spec already documents this. `openapi.json` lists `501` among the responses and the `format` parameter's own description says those four are accepted but unimplemented. The docs page simply did not carry it across.

I kept the four in the contract rather than trimming the list, because accepting them is deliberate: a client written against the documented set gets an honest "not yet" instead of being turned away as though it had sent a typo, and the dashboard shows them disabled rather than hidden for the same reason. Trimming would have hidden the design.

Also notes that `html` runs the other way. It renders on REST and is deliberately **not** one of the six formats the MCP tools offer, because an interactive viewer is no use to an agent. Neither list is a subset of the other, which is worth saying once rather than leaving two pages looking inconsistent.

`mcp.mdx` already handled the general case, so this brings the REST table up to the same standard.

`make validate` passes at 413 pages. The filesystem link check over 2,785 internal links finds only the pre-existing #344.

https://claude.ai/code/session_01T8w822GqnrviVMytnAsKFP


nc9:
Closing: the four formats are being wired into the REST endpoint in squirrelscan/repo#2106 (renderers exist at 97d2a284), so the 501 path goes away before these endpoints ever reach prod. The docs table becomes true rather than caveated.

PR #349, opened 2026-09-13
fix(report): print the requested page limit alongside the effective one

nc9:
`scanScope.requestedMaxPages` has been on the contract since #1909, and the `json` and `llm` outputs already emit it. The three renderers a person actually reads never did.

So a run reduced from 10,000 pages to 4,320 printed `page limit 4320`, which reads as the number the user chose, and the full-scan hint then told them to **raise** a limit that was already above what the run could use.

## What changed

- `scanScopeLine` prints `page limit 4320 of 10000 requested` when the effective limit is smaller than the requested one. Text, markdown and HTML all share it, as does the CLI console. Unclamped output is unchanged byte for byte: the comma only appears where the sentence gains a second clause.
- `fullScanHint` stops advising a raise in that case, in both of its branches, and states what happened instead.

The reduction is detected by **comparing** the two numbers rather than by the presence of `requestedMaxPages` alone, so a producer that echoes the same value both ways cannot render a reduction that never happened.

## Why the wording is neutral

Two different things set this field, and they want different advice:

- the CLI's own page cap, when a request exceeds `MAX_PAGES_CAP`
- a hosted run whose crawl budget cannot pay for the pace a site asks for, where the useful next step is narrowing the crawl

This renderer is shared by both, so it states the facts and leaves the remedy to the surface that knows it. The hosted dashboard says the surface-specific part separately.

## Tests

`packages/report/tests/scan-scope.test.ts` gains 8 cases: both limits in the scan line, the uncapped variant, the CLI reading identically to the cloud, a requested limit equal to the effective one not counting as a reduction, the hint dropping its raise advice in both branches, an unclamped run keeping its original advice unchanged, and all three human-readable renderers carrying both numbers.

379 pass across `packages/report/tests`, plus the CLI suites that consume these two functions (`page-limit`, `publish-slim`). tsgo clean.

`packages/report` has no `format`/`lint` script, so the root `--if-present` scripts skip it; `coverage.ts` is not oxfmt-clean on `main` today and I have deliberately not reformatted it, to keep this diff to the change.

## Follow-up

The private gitlink bump that picks this up lands separately.


nc9:
Ran a local codex pass since this repo has no automatic PR review. Verdict was ship, with no blocking findings. Two things from it worth recording here:

- It diffed **7,776 unclamped output combinations** against the parent commit and found them byte-for-byte identical, which is the property this change most needed to hold. Missing `maxPages` still suppresses the limit, `requested <= effective` keeps the old wording, and a fractional value keeps its digits rather than rounding into a false equality.
- It confirmed no consumer parses or asserts the old strings: text, markdown, HTML, the CLI console and the CLI audit command all just display what these functions return, and the audit command's existing guard still prevents a duplicate hint.

One adjacent gap it noticed, pre-existing and deliberately not touched here: `packages/report/src/output/xml.ts` omits `scanScope` altogether, so the XML renderer discloses neither limit. That predates this change and is a separate parity question from the one this PR closes. Happy to file it if useful.

All 14 checks pass, including the sign-off and secret-scan gates.

PR #351, opened 2026-09-13
docs: cloud crawl scope, include/exclude patterns and authorised crawls (#348)

### LINKED ISSUE #348: docs: cloud crawl scope patterns and the authorised crawl profile
squirrelscan/repo#2118 changed two things a user can see on a website's crawl settings page, and the docs do not describe either.

**Include and exclude patterns now apply to cloud audits.** They were editable and stored before that PR but nothing on the run path read them, so a customer could set exclusions and watch the crawl ignore them. They now reach every cloud audit of that site: manual, scheduled, the initial audit after adding a site, the GitHub quick-check, and an infra retry. A run records the scope it dispatched with, so a retry reproduces the crawl that failed rather than one re-scoped by settings that moved since.

Worth documenting alongside that: the patterns are bounded (50 per list, 500 characters each, 4KB encoded), and the settings page now refuses a list that exceeds those rather than accepting it and dropping the tail at crawl time.

**Authorised crawls.** A website with a custom request header configured for its cloud audits now crawls under an "authorised" profile: higher concurrency, no fixed per-host delay, and robots.txt is not applied. The last part needs saying plainly, because it is wider than it sounds. The engine gates robots `Disallow` and robots `Crawl-delay` on one flag, so an authorised crawl ignores both, and anything a site previously kept out with `Disallow` needs an exclude pattern instead. The crawl still backs off per host on 429/503/`Retry-After`.

The settings page shows the resulting status read-only ("Authorised crawl: on (custom header)") and disables the robots toggle while it is on, keeping the saved preference for when it is off.

Suggested home: the cloud audit / website settings pages, wherever custom request headers are already described.

## Acceptance Criteria
- [ ] Docs state that a website's include and exclude patterns apply to its cloud audits, and which triggers that covers
- [ ] Docs state the pattern limits and that the settings page rejects a list exceeding them
- [ ] Docs describe the authorised crawl profile, what turns it on, and what it changes
- [ ] Docs state plainly that an authorised crawl does not apply robots.txt at all, and that exclude patterns are the way to keep sections out
- [ ] `make validate` passes in `docs/`

nc9:
Closes #348.

Documents the cloud crawl settings, which had no docs page at all, and corrects the include/exclude pattern syntax that the existing config reference got wrong.

## New page: `dashboard/crawl-settings`

Covers the four cards on **Website → Settings → Crawl**:

- **Crawl scope**, `discover` / `sitemap`, from squirrelscan/repo#2109, with the empty-sitemap fallback and the run notice it posts.
- **Include and exclude patterns** (squirrelscan/repo#2108): where they apply, the limits, and that a list breaking one is refused at save time rather than truncated at crawl time.
- **Respect robots.txt**, including that `Crawl-delay` is capped at 2 seconds and what that costs a large crawl.
- **Authorised crawls** (squirrelscan/repo#2113): what turns one on, and plainly that it does not apply robots.txt at all, so an exclude pattern is the replacement for a `Disallow`.

## Correction to `configuration/crawler`

The `include` / `exclude` reference described a glob dialect the crawler does not implement. `matchesPattern` treats `*` as the only special character, standing for any run of characters including `/`, and escapes everything else, so `?` and `[abc]` match themselves. A pattern is tested against the URL's **path** unless it starts with `http://` or `https://`.

Three things followed from that:

- The documented `*?preview=*` examples could never match, because a path has no query string. They are absolute now.
- `/blog/*` was documented as not matching `/blog/post/comment`. It does: a `/*` suffix is a prefix test.
- `/api/*/users` was documented as not matching `/api/v1/v2/users`. It does, for the same reason.

Also adds two cross-reference notes: from `include` to the dashboard lists, and from `headers` to authorised crawls.

## Verification

Every claim on the new page was checked against the implementation, and a second adversarial pass over it found nine factual errors that are fixed in this branch. The ones worth calling out, because they are the non-obvious product behaviour:

- The GitHub check forwards scope and patterns but no custom headers, so it is never an authorised crawl and does not read the robots preference.
- Quick coverage (the **Fast** depth preset) disables link discovery on its own, so **Follow links** behaves like sitemap-only there.
- Sitemap scope still queues pages carrying findings open from the last audit, and the start URL is queued but still filtered like any other URL.
- Backoff is `429`, or `503` **with** a `Retry-After`. A bare `503` is an outage.

`make validate` passes in `docs/`.

## Notes

- The **Crawl scope** section describes squirrelscan/repo#2109, which is not merged yet. Hold this until that ships.
- Filed squirrelscan/repo#2134 for letting the MCP `add_website` tool set a site's crawl scope at registration.
- Not fixed here, flagged instead: `configuration/crawler` documents `allow_query_params` default as "drop all query params", but an empty list drops only known tracking prefixes and keeps the rest.
- The private gitlink bump follows separately.


nc9:
Updated in c94d7c0 for a behaviour change that landed after this PR was opened.

squirrelscan/repo#2132's acceptance pass found that `disableLinkDiscovery` gates only one of the three ways a URL enters the crawl frontier. Two others were letting off-sitemap pages in, and both are now turned off under sitemap scope:

- **Carried-finding seeds.** They are enqueued ahead of the sitemap and win the frontier dedup, so an off-sitemap page spent the capped budget before the URLs the customer asked for. This PR previously said those pages are still queued, which is now false.
- **Cached-page reuse.** Reusing a page re-reads its links, and the crawler enqueues them with no discovery check, so one reused page puts the whole site back in the frontier. Tracked as #354.

Both now sit under a "What sitemap scope gives up" heading, because someone weighing the setting needs the cost and not only the benefit. The sample event log carries the reuse sentence the run actually posts, and the empty-sitemap fallback moved back above the trade-offs so the section reads in the order the crawl happens.

Also added: an authorised crawl never reuses pages either, because an authenticated response must not be stored where another audit could read it.

One claim I removed rather than corrected. An earlier draft tied reuse to the dashboard's **Incremental crawl** toggle. Checking the source, that toggle is stored in the Durable Object and never read on any run path, so reuse is decided entirely by the container's environment and the run's shape. Filed as squirrelscan/repo#2136. The page now describes what the crawl does and does not mention the toggle.

`make validate` still passes. All checks green.

PR #388, opened 2026-09-19
fix(release): sign the darwin binaries after compiling them

nc9:
Fixes the invalid ad-hoc signature on the darwin release binaries (squirrelscan/repo#2310).

## The bug

`bun build --compile` writes its bundle into the Mach-O and moves the linker's ad-hoc signature to the end of the file, but it never recomputes the page hashes that signature seals. Every darwin asset we have published therefore fails strict verification, and a Mac that enforces signatures kills the binary before it runs a line. That is the exit 137 at `verify_binary_killed` that Apple Silicon installs have been reporting.

Reproduced on an arm64 Mac against a binary built exactly the way the release builds it:

```
$ codesign -dv squirrel-0.0.96-darwin-arm64
CodeDirectory v=20400 size=559390 flags=0x20002(adhoc,linker-signed) hashes=17478+0
$ codesign --verify --strict squirrel-0.0.96-darwin-arm64
squirrel-0.0.96-darwin-arm64: invalid signature (code or signature have been modified)
```

After `make sign-darwin`:

```
$ codesign -dv squirrel-0.0.96-darwin-arm64
Identifier=com.squirrelscan.squirrel
CodeDirectory v=20400 size=559474 flags=0x2(adhoc) hashes=17478+2
$ codesign --verify --strict squirrel-0.0.96-darwin-arm64 ; echo $?
0
$ ./squirrel-0.0.96-darwin-arm64 --version
0.0.96
```

## What changed

**`scripts/sign-darwin.sh`** re-signs both darwin binaries ad-hoc. No certificate is involved: re-signing rehashes the finished file, which is the whole fix. It uses `rcodesign` when it is on PATH (the release, which builds on Linux) and Apple's `codesign` otherwise (a Mac). The identifier is pinned to `com.squirrelscan.squirrel` rather than defaulted from the filename, which carries the version, so an allowlist rule in an MDM or endpoint-security policy does not stop matching at every release.

Both tools were checked against Apple's own verifier on this machine: a binary signed by `rcodesign` on the Linux code path passes `codesign --verify --strict` and runs.

**`make build-all`** ends with `make sign-darwin`, so the release and a local build take the same path, and signing always happens before `make manifest`. Signing changes the file bytes: a manifest generated first records hashes no published asset matches, and every `squirrel self update` would then fail its checksum check. Measured on the same binary: `583d2eed...` before signing, `d7ef7f20...` after.

**The release workflow** installs `rcodesign` 0.29.0 pinned by version and by sha256 (never `cargo install` at release time), and gains a `verify-darwin` job on `macos-latest` that `github-release` now depends on. It runs Apple's verifier on both darwin assets, re-hashes them against `manifest.json`, and runs the arm64 binary, all before anything publishes. `npm`, `installers-publish` and the R2 channel metadata all sit behind `github-release`, so nothing reaches users past a failed check.

**The macOS install-test legs** run `codesign --verify --strict` on the binary as it lands on disk. That is after publish, so it is a second proof rather than a gate.

**`install.sh`**: on macOS a killed binary no longer reads as an out-of-memory kill. `squirrel --version` does not exhaust a Mac, so the message now names the two causes that are actually possible, a rejected signature or security software, gives the one command that tells them apart, and carries the macOS version. The worker keeps a fixed set of report fields, so the version rides in the message itself, which is what Sentry shows as `error_line`; the Linux 137 guidance is unchanged.

## What the gates can and cannot catch

- The `verify-darwin` job is authoritative: it is Apple's verifier, on a real Mac, on the exact bytes the release is about to upload, and it also proves the manifest was generated from the signed bytes.
- The Linux-side check inside `sign-darwin.sh` is not. `rcodesign verify` says itself that it is unreliable and it reports a valid ad-hoc signature as a CMS failure, so its exit code cannot be used. The script allows only the complaints an ad-hoc signature is expected to draw and fails on anything else, including an empty or unrecognised output, so a crashed or confused verifier fails the build rather than passing it.
- Nothing here signs with a Developer ID or notarises. The binaries stay ad-hoc signed, which is what makes them runnable, not what makes Gatekeeper trust a download.

## Tests

`scripts/install-contract.test.ts`, 114 pass / 0 fail. Six new tests: the macOS guidance, the Linux guidance, the reported macOS version, the real `is_darwin` and `macos_product_version` driven through shimmed `uname` and `sw_vers`, a Mac where `sw_vers` fails, and the quoting of a path with a space.

Mutation-tested, each confirmed red then restored: dropping the darwin branch, dropping the `is_darwin` guard, dropping the macOS version from the report, no-op signing, a verifier that crashes, and a verifier that says nothing.


nc9:
`Dependency audit` is red on this PR and is not caused by it. `bun audit` reads live advisory data, and GHSA-9rgm-9g3h-6x36 (devalue < 5.9.1, reached through `docs > tangly > astro`) was published after main's last CI run on 2026-09-17. This branch touches no dependency file: `bun.lock` and every `package.json` are untouched. Tracked separately in squirrelscan/repo#2317; every other check is green.

PR #399, opened 2026-09-22
fix(entity-map): publish a type-stratified sample instead of an alphabetical slab

nc9:
Stacks on #398 (base `fix/entity-map-scale-bounds`), so the diff here is the sampler only.

## The bug

`slimEntityMapForPublish` ranked nodes by `occurrences` and broke ties on `key`. A store that declares its entities per page (Shopify, WooCommerce, most of the CMS world) gives every Product, Offer, BreadcrumbList and WebPage `occurrences === 1`, so the ranking fell through to its tie-break and the published body became an alphabetical slice of the catalogue: measured on a synthetic Shopify-shaped map, 125 Products, 125 BreadcrumbLists, 124 WebPages, **zero Offers**, identical at 500, 1,000 and 5,000 pages. Half the body went to page-local entities the graph hides by default.

The byte backstop compounded it: on the first overshoot it halved the node budget, so a 5,000-page map published 375 nodes and 266KB against a 512KB gate.

## The fix

Three tiers, stratified by type inside each one:

1. **shared subjects** (declared on more than one page) first, by occurrences then degree
2. **one-off subjects**, round-robin across types so a 5,000-product catalogue cannot crowd out its own Offers; inside a type by degree (inbound edges included, since an Offer is only ever pointed at)
3. **page-local** entities last, capped at `maxPageLocalShare` of the budget (publish 0.1, viewer 0.25), with the unused slice handed back if the other tiers cannot fill it

The byte loop now scales the budget to the overshoot instead of halving it. Shrinking the budget shrinks every tier and bucket with it, so the mix survives and only the resolution drops. `summary` is still never recomputed; `truncated` still reconciles.

## Before / after

Synthetic Shopify-shaped map (Product + Organization + BreadcrumbList + WebPage + nested Offer per page), published body:

| pages | full nodes | | published nodes | edges | types | non-page-local | bytes |
|---|---|---|---|---|---|---|---|
| 500 | 2,001 | before | 375 | 373 | 4 | 126 | 266,027 |
| | | **after** | **677** | **677** | **5** | **610** | **511,678** |
| 1,000 | 4,001 | before | 375 | 373 | 4 | 126 | 266,036 |
| | | **after** | **677** | **677** | **5** | **610** | **511,687** |
| 5,000 | 20,001 | before | 375 | 373 | 4 | 126 | 269,562 |
| | | **after** | **668** | **667** | **5** | **602** | **511,595** |

By type at 1,000 pages: before `BreadcrumbList 125, Product 125, WebPage 124, Organization 1`; after `Product 305, Offer 304, WebPage 34, BreadcrumbList 33, Organization 1`. Projection cost at 5,000 pages is 13ms.

## Tests

`packages/audit-engine/tests/entity-map-slim.test.ts`, fixture built through the real `createEntityMapBuilder` over 1,000 pages of store JSON-LD. Self-reviewed by mutation (codex is out of credits); all five caught:

| mutation | caught by |
|---|---|
| revert to the `occurrences \|\| key` ranking | 5 tests |
| drop the per-type stratification | 5 tests |
| drop the page-local tier | page-local cap |
| halve the budget on overshoot | byte-scaling test |
| count only outbound edges as degree | inbound-degree test |

The first pass of the byte test was **inert**: the 1,000-page fixture fits 750 nodes under the gate, so the loop never ran and the halving mutation survived. Replaced with a fixture sized to overshoot.

## Notes

- The API does not enforce caps on ingest: the TypeBox request schema has no `maxItems`, so the bound is only ever what the CLI chose to send. Private-repo fix, not touched here.
- `docs/entity-map/format.mdx` gains a paragraph on how the sample is chosen; `make validate` passes.


nc9:
---

## Update: both store shapes covered (26d04cc)

Per-page entities come in two flavours and only one of them was in the fixture. Measured on the parent (#398) code, 1,000 pages, same fixture with and without an `@id` on the BreadcrumbList and WebPage:

| shape | | nodes | edges | page-local | composition | bytes |
|---|---|---|---|---|---|---|
| page-locals as blank nodes | before | 750 | **0** | 748 | BreadcrumbList 748, Brand 1, Organization 1 | 184,627 |
| | **after** | 750 | **711** | 75 | Product 337, Offer 336, WebPage 38, BreadcrumbList 37, Brand 1, Organization 1 | 486,117 |
| every entity has an `@id` | before | 750 | 498 | 499 | BreadcrumbList 250, Product 249, WebPage 249, Brand 1, Organization 1 | 426,035 |
| | **after** | 750 | 711 | 75 | Product 337, Offer 336, WebPage 38, BreadcrumbList 37, Brand 1, Organization 1 | 491,141 |

Without an `@id` the page-local nodes take `anon:`/`blank:` keys, which sort ahead of every `id:` key, so they took the entire budget: 750 disconnected breadcrumbs and **not one edge**. With an `@id` the keys group by page URL, the types interleave, and the body looks like a fair sample while being the alphabetically-first pages with zero Offers. Both are now fixture-parameterized in `entity-map-slim.test.ts` (one `blankLocals` flag, every composition assertion runs twice, 36 tests). Mutation round re-run on the wider suite: the alphabetical revert now fails 11 tests, dropping stratification 10, page-local tier 2, halving 1, outbound-only degree 1.

